PR #1, opened 2017-06-28
Add tests & fixes based on them

mpapis:


pluff:
we have 100% coverage now 👍

PR #2, opened 2017-06-28
Configurable dump name

mpapis:
based on #1 so we can test better


pluff:
Closed in favor of https://github.com/toptal/crystalball/pull/19

PR #7, opened 2017-12-07
[CB-5] Add specs to predictor

andriusch:


spec/predictor_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+describe Crystalball::Predictor do
+  subject { described_class.new(map, diff) }
+  let(:repository) { Git::Base.new }
+  let(:file_diff1) { Crystalball::SourceDiff::FileDiff.new(repository, Git::Diff::DiffFile.new(repository, path: 'file1.rb')) }
+  let(:diff) { [file_diff1] }
+  let(:map) { {spec_file: %w[file1.rb]} }
+
+  describe '#cases' do
+    subject { super().cases }
```
pluff: that `super()` notation is a bit weird for me because you change the subject completely. It makes sense when you modify the parent subject a bit but when you provide completely different Object it looks awkward. I'd rather use `subject { predictor.cases }` in this case

pluff (approved):

PR #8, opened 2017-12-08
[CB-7] Add spec coverage with simplecov and code climate

andriusch:
Codeclimate doesn't display reports in PR, although with debug output enabled it shows that everything is working successfully. Might be due to the fact that we don't have coverage for master and it doesn't have anything to compare to. Will investigate after we merge this.


pluff (approved):
please lets investigate why we don't see coverage stats in PRs

PR #12, opened 2017-12-12
Cover MapGenerator with specs

pluff:


pluff:
@andriusch for some reason test coverage information is not present in PR checks

P.S. Also README badges seem to be broken https://github.com/toptal/crystalball

spec/map_generator_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      subject.finalize!
+    end
+  end
+
+  describe '#refresh_for_case' do
+    let(:rspec_example) { double(run: true) }
+    let(:case_map) { instance_double('Crystalball::CaseMap') }
+
+    before do
+      before = {}
```
andriusch: I think we should add some keys to before and after, otherwise you can use them in different order and spec will still pass

andriusch (changes requested):

andriusch (approved):

PR #13, opened 2017-12-12
Add first integration specs

andriusch:


features/local_diff_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+    RUBY
+
+    class2_path.open('w') { |f| f.write <<~RUBY }
+      class Class2
+        def initialize
+          @var = 2
+        end
+      end
+    RUBY
```
pluff: Maybe we can move these 2 initial files + spec file to fixtures?

andriusch: We could have them in fixtures and then copy them to some temporary folder where we would initialize git repo and do all the other things. Not sure which solution would be better current or that one.

pluff (approved):

PR #15, opened 2017-12-15
[CB-13] Predictor improvements

andriusch:


spec/support/shared_examples/simple_git_repository.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+shared_context 'simple git repository' do
+  subject(:forecast) { Crystalball.foresee(workdir: root, map_path: root.join('execution_map.yml')) }
```
pluff: I think this subject is not related to the context.

lib/crystalball/predictor.rb
```diff
@@ -3,17 +3,28 @@
...
 
-    def initialize(map, source_diff)
+    def initialize(map, source_diff, spec_pattern)
       @map = map
       @diff = source_diff
+      @spec_pattern = spec_pattern
     end
 
     def cases
+      cases_from_changed_code + cases_from_changed_specs
```
pluff: I was thinking about it earlier. It feels like there should be an array of prediction patterns that you can pass to predictor. That will make it somewhat extendable.
Something like that:
```
p = Predictor.new(map, source_diff) # by default it has only one default "pattern" which is `cases_from_changed_code`
p.prediction_patterns << Patterns::NewSpecs.new(spec_regexp) # your "new_specs" pattern
p.prediction_patterns << Patterns::CorrespondingRailsFile.new # some other pattern
p.cases
```

pluff (commented):
is there a way to tell if a file was renamed or modified with current functionality? I think `FileDiff` will tell us that it's modified. Renaming status might be useful in predictions

andriusch:
No git gem doesn't show renaming. It shows that one file was removed and another was added, even if `git status` shows that it was renamed.

lib/crystalball/predictor.rb
andriusch: If we do it like that, I'd much rather not add any defaults and allow everything to be specified explicitly. Also how would we use it from `Crystallball.foresee`? Something like:
```ruby
Crystallball.foresee do |predictor|
  predictor.patterns << CrystallBall::Pattern::ChangedDependencies.new
  predictor.patterns << CrystallBall::Pattern::ChangedSpecFiles.new(spec_regexp)
end
```

pluff: yeah something like that

pluff (approved):

PR #16, opened 2017-12-18
[CB-11] Store commit information in execution map file

pluff:


lib/crystalball/git_repo.rb
```diff
@@ -0,0 +1,37 @@
...
+  # Wrapper class representing Git repository
+  class GitRepo
+    attr_reader :repo_path
+
+    def initialize(repo_path)
+      @repo_path = repo_path
+    end
+
+    def source_diff
+      @source_diff ||= SourceDiff.new(self)
```
andriusch: I don't like that we have circular dependency here and it's only used in `pristine?` method, so I think it would be better to remove this method and just initialize `SourceDiff` in 2 places that we need.

lib/crystalball/map_generator/simple_map.rb
```diff
@@ -4,30 +4,57 @@ module Crystalball
...
-      def load
-        self.raw_map = storage.load
+      attr_reader :cases, :metadata
+
+      delegate %i[commit commit=] => :metadata
+
+      def initialize(storage, metadata: nil, cases: {})
+        @storage = storage
+        @cases = cases
+        @metadata = Metadata.new(metadata || {type: self.class.name})
```
andriusch: I think it should be `Metadata.new(type: self.class.name, **metadata)`, so type can be overriden, but it still should always be present, no?

lib/crystalball/map_generator/simple_map.rb
```diff
@@ -4,30 +4,57 @@ module Crystalball
...
       end
 
       def dump
-        storage.dump raw_map
+        storage.dump self
       end
 
-      def clear!
-        self.raw_map = {}
+      def reset!
```
andriusch: I don't like that we have `clear!` and `reset!`, both have pretty much the same meaning, and you can't tell the difference without checking source.

lib/crystalball/map_generator/simple_map.rb
andriusch: Actually do we really need metadata object? It's 2 fields only and complicates things.

lib/crystalball/map_generator/simple_map.rb
```diff
@@ -4,30 +4,57 @@ module Crystalball
...
+        clear!
+        self.metadata = Metadata.new(type: self.class.name)
+      end
+
+      def serialize
```
andriusch: `def to_h`

lib/crystalball/map_generator/simple_map.rb
```diff
@@ -4,30 +4,57 @@ module Crystalball
...
+        @cases = cases
+        @metadata = Metadata.new(metadata || {type: self.class.name})
       end
 
       def stash(case_map)
-        raw_map[case_map.case_uid] = case_map.coverage
+        cases[case_map.case_uid] = case_map.coverage
       end
 
       def dump
```
andriusch, line 37: This `dump` seems outside of scope of `Map` class which aside from this method is just a data class. I get that it's needed for `StandardMap`, but I think we've been thinking about it wrong. Instead of `SimpleMap` and `StandardMap`, we should actually have one `Map` class, and `SimpleMapGenerator` and `StandardMapGenerator`.

lib/crystalball/map_generator/standard_map.rb
```diff
@@ -4,22 +4,34 @@ module Crystalball
   class MapGenerator
     # Simple map storing object with threshold for dumping to persistent storage
     class StandardMap < SimpleMap
-      def initialize(*args, dump_threshold: 100)
-        super(*args)
-        @dump_threshold = dump_threshold
+      def initialize(*args, **options)
```
andriusch: `(*args, dump_threshold: 100, **options)`

lib/crystalball/map_storage/yaml_storage.rb
```diff
@@ -17,11 +17,19 @@ def clear!
...
+        cases = cases.inject(&:merge)
+
+        Object.const_get(metadata[:type]).new(self, metadata: metadata, cases: cases)
       end
 
-      def dump(map)
-        path.open('a') { |f| f.write YAML.dump(map) }
+      def dump(map, exclude_metadata: false)
+        path.open('a') do |f|
+          f.write YAML.dump(map.serialize[:metadata]) unless exclude_metadata
```
andriusch: I don't think it should be responsibility responsibility of `YAMLStorage` when to store what. Rather we should have 2 methods `dump_metadata` and `dump_cases` and let storage engines only decide how do we store the data, not when do we store the data.

lib/crystalball/map_storage/yaml_storage.rb
```diff
@@ -17,11 +17,19 @@ def clear!
       end
 
       def load
-        YAML.safe_load(path.read) if path.exist?
+        metadata, *cases = path.read.split("---\n").reject(&:empty?).map do |yaml|
+          YAML.safe_load(yaml, [Symbol])
+        end
+        cases = cases.inject(&:merge)
```
andriusch: `merge!`?

andriusch (changes requested):

lib/crystalball/map_storage/yaml_storage.rb
pluff: YAMLStorage doesn't decide. StandardMap decides :)

lib/crystalball/map_generator/simple_map.rb
pluff: I think we need a separate ticket for that. I'll create one

lib/crystalball/map_generator/simple_map.rb
pluff: Well its a nice interface where to store your non-case related information. If user needs to store something else they can easily extend their own Metadata class.

lib/crystalball/git_repo.rb
pluff: its used in `pristine?` for now but we'll use it later for other extensions to git repo.

andriusch (approved):

PR #17, opened 2017-12-19
Move out storage logic to a generator

pluff:


lib/crystalball/map_generator.rb
```diff
@@ -28,48 +28,58 @@ def build(config)
...
-    attr_reader :execution_detector, :map_storage, :map_class
+    attr_reader :execution_detector, :map_storage, :dump_threshold
     attr_writer :map
 
     def repo
       @repo ||= GitRepo.new('.')
     end
+
+    def check_dump_threshold
+      return unless dump_threshold.positive? && map.size >= dump_threshold
```
andriusch, line 79: I think `dump_storage: nil` would be better when user doesn't want dump storage, rather than setting it to 0 or -1

andriusch (approved):

lib/crystalball/map_generator.rb
pluff: `nil` will also work since we do `.to_i`

PR #19, opened 2017-12-20
[CB-14] Allow configuring map path

andriusch:


spec/map_generator/configuration_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  it 'converts dump threshold to integer' do
+    subject.dump_threshold = '10'
+    expect(subject.dump_threshold).to eq(10)
+  end
+
+  it 'converts map storage path to pathname' do
+    subject.map_storage_path = 'my_map.yml'
+    expect(subject.map_storage_path).to eq(Pathname('my_map.yml'))
+  end
+end
```
pluff: can we follow common spec structure here? 
`describe '#method' { }`

lib/crystalball/map_generator/configuration.rb
```diff
@@ -0,0 +1,35 @@
...
+    # `Crystalball::MapGenerator.start! { |config| config } block.
+    class Configuration
+      attr_writer :execution_detector, :map_storage
+
+      def execution_detector
+        @execution_detector ||= ExecutionDetector.new(Dir.pwd)
+      end
+
+      def map_storage_path
+        @map_storage_path ||= Pathname('execution_map.yml')
```
pluff, line 15: I wonder if we really need this default value :)

pluff (approved):

lib/crystalball/map_generator/configuration.rb
andriusch: I'd guess that most users wouldn't care enough to customize it, so it's always better to have some default for configuration

PR #21, opened 2018-01-10
[CB-18] Allow YAMLStorage to receive directory of map files

pluff:


lib/crystalball/map_storage/yaml_storage.rb
```diff
@@ -8,6 +8,41 @@ class MapStorage
...
+              YAML.safe_load(yaml, [Symbol])
+            end
+            cases = cases.inject(&:merge!)
+
+            [metadata, cases]
+          end
+        end
+
+        def guard_metadata_consistency(data)
+          commits = data.map { |d| d.first[:commit] }.uniq
```
andriusch: I think only checking commit and type is not future proof. Instead we should check that every all metadata is same.

lib/crystalball/map_storage/yaml_storage.rb
```diff
@@ -8,6 +8,41 @@ class MapStorage
...
+            metadata, *cases = file.read.split("---\n").reject(&:empty?).map do |yaml|
+              YAML.safe_load(yaml, [Symbol])
+            end
+            cases = cases.inject(&:merge!)
+
+            [metadata, cases]
+          end
+        end
+
+        def guard_metadata_consistency(data)
```
andriusch: We should pass only metadata to this method.

andriusch (changes requested):

lib/crystalball/map_storage/yaml_storage.rb
```diff
@@ -8,6 +8,38 @@ class MapStorage
     class YAMLStorage
       attr_reader :path
 
+      class << self
+        def load(path)
+          ymls = read_files(path)
+          meta = ymls.map(&:first)
```
andriusch: `meta, cases = *ymls.transpose`

andriusch (approved):

PR #24, opened 2018-01-16
[CB-22] Increase coverage

EPecherkin:
[CB-22](https://youtrack.toptal.net/youtrack/issue/CB-22)

Exclude spec/support/shared_examples/ from coverage
Add new features for better check

`Coverage report generated for RSpec to /work/toptal/crystalball/coverage. 614 / 614 LOC (100.0%) covered.`


lib/crystalball/source_diff.rb
```diff
@@ -28,7 +28,18 @@ def empty?
...
+        # FIXME: ruby-git marks moved files as modified. It works for lib files, but not for specs
+        # This code splits diff of moved file to two separate diffs with removed and added file
+        if diff.moved?
+          added_file_diff = FileDiff.new(file_diff, override_type: 'new', override_relative_path: diff.new_relative_path)
+          removed_file_diff = FileDiff.new(file_diff, override_type: 'deleted')
+          [removed_file_diff, added_file_diff]
+        else
+          diff
+        end
+      end.flatten
```
pluff: this looks weird. can you elaborate more on it?

pluff (changes requested):

lib/crystalball/source_diff.rb
EPecherkin: Git will mark a file as `modified` when you rename it. It works well for source files. But not for spec files. It will try to run a file with the old name and completely ignores the new file.
If spec file was renamed, it means that previous filename should be removed from prediction map and the new one should be added.
I do it manually. If patch indicates that file was moved, I manually create a diff with deleted file and with the added one

EPecherkin: done

lib/crystalball/source_diff/file_diff.rb
```diff
@@ -4,24 +4,34 @@ module Crystalball
...
     # Data object for single file in Git repo diff
     class FileDiff
-      def initialize(git_diff)
+      attr_reader :relative_path
+
+      def initialize(git_diff, override_type: git_diff.type, override_relative_path: git_diff.path)
         @git_diff = git_diff
+        @type = override_type
+        @relative_path = override_relative_path
+      end
```
pluff: add specs and cleanup code

pluff (approved):

PR #28, opened 2018-01-23
[CB-23] Predict failures like changed validations of a model

EPecherkin:
Add new DirectMapPredictor
    
A developer is able to specify regexp to match specific source files to specific spec files. It will add proper spec files to a prediction set if matched source files were changed

```ruby
predictor.use Crystalball::Predictor::DirectMapPredictor.new from: /lib\/models\/(.*).rb/,
                                                             to: './spec/models/%s_spec.rb'
```

Should be merged after https://github.com/toptal/crystalball/pull/24


features/fixtures/simple_app/spec/class1_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'spec_helper'
 
 describe Class1 do
-  subject { Class1.new }
+  subject { described_class.new }
```
pluff: this is default subject

features/fixtures/simple_app/spec/class2_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'spec_helper'
 
 describe Class2 do
-  subject { Class2.new }
+  subject { described_class.new }
```
pluff: same as above

features/fixtures/simple_app/spec/models/model1_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Model1 do
+  subject { described_class.new }
```
pluff: same as above

lib/crystalball/predictor.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'crystalball/predictor/modified_execution_paths'
 require 'crystalball/predictor/modified_specs'
+require 'crystalball/predictor/direct_map'
```
pluff: predictor class by itself doesn't depend on any of strategies required above. Let's move those requires to the place where we use them

lib/crystalball/predictor/direct_map.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Crystalball
+  class Predictor
+    # Used with `predictor.use Crystalball::Predictor::DirectMap.new(from: %r{models/(.*).rb}, to: "./spec/models/%s_spec.rb")`.
+    # When used will look for files matched to `from` regex and use captures to fill `to` string to
+    # get paths of proper specs
+    class DirectMapPredictor
```
pluff: its not a predictor. It's a prediction strategy. `Crystalball::Predictor` is a predictor. Please rename a class to something like `AssociatedSpecs`

spec/predictor/direct_map_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Crystalball::Predictor::DirectMapPredictor do
+  subject(:predictor) { described_class.new from: %r{models/(?<file>.*).rb}, to: 'spec/%<file>s_spec.rb' }
+  let(:repository) { Git::Base.new }
+  let(:path1) { 'models/file1.rb' }
+  let(:spec_path1) { 'spec/file1_spec.rb' }
+  let(:file_diff1) { Crystalball::SourceDiff::FileDiff.new(Git::Diff::DiffFile.new(repository, path: path1)) }
```
pluff: you can mock it to be double with `relative_path` stub

features/change_files_spec.rb
```diff
@@ -6,6 +6,8 @@
   subject(:forecast) do
     Crystalball.foresee(workdir: root, map_path: root.join('execution_map.yml')) do |predictor|
       predictor.use Crystalball::Predictor::ModifiedExecutionPaths.new
+      predictor.use Crystalball::Predictor::DirectMapPredictor.new from: %r{lib/models/(?<file>.*).rb},
+                                                                   to: './spec/models/%<file>s_spec.rb'
```
pluff: maybe it worth to have separate feature file per each prediction strategy

pluff (changes requested):

lib/crystalball/source_diff.rb
```diff
@@ -28,7 +28,18 @@ def empty?
 
     # TODO: Include untracked to changeset
     def changeset
-      @changeset ||= git_diff.map { |file_diff| FileDiff.new(file_diff) }
+      @changeset ||= git_diff.map do |file_diff|
+        diff = FileDiff.new(file_diff)
+        # FIXME: ruby-git marks moved files as modified. It works for lib files, but not for specs
```
andriusch: I remember testing it and it was the opposite. `ruby-git` instead of marking files as moved, would mark them as old file removed, new file added.

lib/crystalball/source_diff.rb
```diff
@@ -28,7 +28,18 @@ def empty?
 
     # TODO: Include untracked to changeset
     def changeset
-      @changeset ||= git_diff.map { |file_diff| FileDiff.new(file_diff) }
+      @changeset ||= git_diff.map do |file_diff|
```
andriusch: You could just use `flat_map`

lib/crystalball/source_diff/file_diff.rb
```diff
@@ -4,24 +4,34 @@ module Crystalball
   class SourceDiff
     # Data object for single file in Git repo diff
     class FileDiff
-      def initialize(git_diff)
+      attr_reader :relative_path
+
+      def initialize(git_diff, override_type: git_diff.type, override_relative_path: git_diff.path)
```
andriusch: I think it would be better to just name them `type:` and `relative_path:`

features/change_files_spec.rb
EPecherkin: makes sense

lib/crystalball/predictor.rb
EPecherkin: It feels like these predictors should be available after `require 'crystalball'`. And this file is good to include everything related to predictors, including predictors themselves

lib/crystalball/predictor/direct_map.rb
EPecherkin: And to use it with the same interface as `Predictor`. Hmm. Maybe we need to rename it
I don't like an idea to have different conceptions to do the similar staff

lib/crystalball/source_diff.rb
EPecherkin: hm. a difference of versions maybe? I would like to test it if possible. On my machine, it is `modified`

lib/crystalball/source_diff.rb
EPecherkin: Right! But some specs failed :thinking: I need to explore it a bit

lib/crystalball/predictor.rb
EPecherkin: done

lib/crystalball/predictor/direct_map.rb
EPecherkin: done

lib/crystalball/source_diff.rb
EPecherkin: no need for this flat_map anymore

spec/predictor/associated_specs_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Crystalball::Predictor::AssociatedSpecs do
+  subject(:predictor) { described_class.new from: %r{models/(?<file>.*).rb}, to: 'spec/%<file>s_spec.rb' }
+  let(:repository) { Git::Base.new }
```
pluff: this one is not used

pluff (approved):

PR #30, opened 2018-01-30
[CB-26] Predict existing specs files only

EPecherkin:
[CB-26](https://youtrack.toptal.net/youtrack/issue/CB-26)

### DIFF
--- a/lib/crystalball/predictor.rb
+++ b/lib/crystalball/predictor.rb
@@ -18,7 +18,19 @@ module Crystalball
 
     # TODO: check if it would be better to call predictors with one case instead of passing the whole map.
     def cases
-      predictors.flat_map { |predictor| predictor.call(diff, map) }
+      # TODO: minimize cases? like [./some/spec1, ./some/spec2, ./some/] -> [./some/]
+      predictors
+        .flat_map { |predictor| predictor.call(diff, map) }
+        .compact
+        .select { |example| File.exist?(example_to_file_path(example)) }
+        .uniq
+    end
+
+    private
+
+    def example_to_file_path(example)
+      root = diff.repository.dir.path
+      File.join(root, example).split('[').first
     end
   end
 end
--- a/lib/crystalball/predictor/associated_specs.rb
+++ b/lib/crystalball/predictor/associated_specs.rb
@@ -12,9 +12,8 @@ module Crystalball
       end
 
       def call(diff, _)
-        diff.map(&:relative_path).grep(from).map do |source_file_path|
-          to % captures(source_file_path)
-        end
+        diff.map(&:relative_path).grep(from)
+            .map { |source_file_path| to % captures(source_file_path) }
       end
 
       private
--- a/lib/crystalball/predictor_evaluator.rb
+++ b/lib/crystalball/predictor_evaluator.rb
@@ -37,8 +37,7 @@ module Crystalball
     end
 
     def prediction_size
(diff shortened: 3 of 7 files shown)


lib/crystalball/predictor/associated_specs.rb
```diff
@@ -12,9 +12,9 @@ def initialize(from:, to:)
       end
 
       def call(diff, _)
-        diff.map(&:relative_path).grep(from).map do |source_file_path|
-          to % captures(source_file_path)
-        end
+        diff.map(&:relative_path).grep(from)
+            .map { |source_file_path| to % captures(source_file_path) }
+            .select { |spec_file_path| File.exist?(spec_file_path) }
```
pluff: I think it's better to move this check to upper level `Predictor#cases` since any of strategies can return specs which don't exist in latest master. You'll also have to cut off `[1:1:2:3]` part

pluff: + we need a feature for it

lib/crystalball/predictor.rb
```diff
@@ -18,7 +18,20 @@ def use(predictor)
...
+        .select do |example|
+          File.exist?(example_to_file_path(example))
+        end
+    end
+
+    private
+
+    def example_to_file_path(example)
+      root = diff.repository.dir.path
+      File.join(root, example).match(/([^\[\]]*)(?:\[.*\])?/)[1]
```
pluff: `File.join(root, example).split('[').first`

lib/crystalball/predictor/associated_specs.rb
```diff
@@ -6,15 +6,16 @@ class Predictor
     # When used will look for files matched to `from` regex and use captures to fill `to` string to
     # get paths of proper specs
     class AssociatedSpecs
+      ALL = './'
+
```
pluff: not used

lib/crystalball/predictor_evaluator.rb
```diff
@@ -37,6 +37,7 @@ def prediction_rate
     end
 
     def prediction_size
+      # FIXME: doesn't work if contains folder prediction. need to expand folder to example to count it properly
```
pluff: something like 
`predictor.map.cases.keys.select { |example| prediction.any? { |p| example.include?(p) } }.size`

lib/crystalball/predictor.rb
EPecherkin: too simple :D

lib/crystalball/predictor/associated_specs.rb
EPecherkin: I'm going to use it here https://github.com/toptal/crystalball-toolbox/pull/3/files#diff-b8ee0ab8664c89405a7e92af9c552629R17
and maybe to include all examples if we meet such thing in prediction list

pluff: I believe it's project specific. some projects require `./` others `./spec` others `./test` etc. So it shouldn't be here

spec/source_diff_spec.rb
```diff
@@ -31,4 +31,28 @@
       it { is_expected.to be_empty }
     end
   end
+
+  describe '#repository' do
+    subject { super().repository }
```
pluff: !

pluff (approved):

PR #31, opened 2018-02-01
Add support of multiple strategies for MapGenerator

pluff:


lib/crystalball/map_generator.rb
```diff
@@ -39,19 +34,17 @@ def start!
       self.map = nil
       map_storage.clear!
       map_storage.dump(map.metadata.to_h)
+
+      strategies.each(&:after_start)
     end
 
     def refresh_for_case(example)
```
EPecherkin, line 41: `flush_for_case`?

EPecherkin (approved):

PR #34, opened 2018-02-02
[CB-24] Add views to execution_map

EPecherkin:
[CB-24](https://youtrack.toptal.net/youtrack/issue/CB-24)


lib/crystalball/map_generator/action_view_strategy.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'coverage'
```
pluff: not needed

lib/crystalball/map_generator/action_view_strategy.rb
```diff
@@ -0,0 +1,37 @@
...
+
+module Crystalball
+  # Class for generating execution map during RSpec build execution
+  class MapGenerator
+    # Map generator strategy to build map of views affected by an example.
+    # It patches ActionView::Renderer.render to get original name of compiled views.
+    class ActionViewStrategy
```
pluff: please nest it under `Crystalball::MapGenerator::Rails` namespace so we have all rails-specific stuff separated

lib/crystalball.rb
```diff
@@ -10,6 +10,7 @@
 require 'crystalball/map_generator'
 require 'crystalball/map_generator/configuration'
 require 'crystalball/map_generator/coverage_strategy'
+require 'crystalball/rails/map_generator/action_view_strategy'
```
pluff: this shouldn't be require by default. It should be required in `crystalball/rails` only

lib/crystalball/rails/map_generator/action_view_strategy/patch.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Crystalball
+  module Rails
+    class MapGenerator
+      class ActionViewStrategy
+        # Module to add new patched `compile!` method to ActionView::Template
+        module Patch
+          def new_compile!(mod)
```
pluff: notation is `def compile!(view)`. please rename the param

lib/crystalball.rb
pluff: this should be required in `crystalball/rails` only. If user wants to use Crystalball with rails he'll have to require `crystalball/rails` explicitly to get rails-specific features.

lib/crystalball/rails/map_generator/action_view_strategy.rb
```diff
@@ -0,0 +1,52 @@
...
+          ::ActionView::Template.class_eval do
+            include Crystalball::Rails::MapGenerator::ActionViewStrategy::Patch
+            alias_method :old_compile!, :compile!
+            alias_method :compile!, :new_compile!
+          end
+        end
+
+        def before_finalize
+          ::ActionView::Template.class_eval do
+            alias_method :compile!, :old_compile! # rubocop:disable Lint/DuplicateMethods
```
pluff: undefine `:new_compile!`?

pluff: and maybe bundle all the magic together?
Something like this would be nice:
```
def after_start
  Patch.apply!
end

def before_finalize
  Patch.revert!
end
```

lib/crystalball/rails/map_generator/action_view_strategy/patch.rb
```diff
@@ -0,0 +1,17 @@
...
+            Crystalball::Rails::MapGenerator::ActionViewStrategy.views.push identifier
```
pluff: simple `ActionViewStrategy` should work here

pluff (changes requested):

pluff (approved):

PR #37, opened 2018-02-07
Allow map class to be configured

pluff:


lib/crystalball/case_map.rb
```diff
@@ -3,20 +3,15 @@
...
-    attr_reader :uid, :affected_files
+    attr_reader :uid, :file_path, :affected_files
     extend Forwardable
 
     delegate %i[push] => :affected_files
 
     def initialize(example, affected_files = [])
-      @uid = build_uid(example)
+      @uid = example.id
+      @file_path = example.file_path
```
EPecherkin, line 13: `delegate :file_path, to: :example` ?

EPecherkin (approved):

lib/crystalball/case_map.rb
pluff: no. I don't want to save the reference to example object to avoid memory bloating

PR #38, opened 2018-02-08
[CB-35] Check file exsits in revision, not on file system

EPecherkin:
[CB-35](https://youtrack.toptal.net/youtrack/issue/CB-35)


pluff:
No

PR #40, opened 2018-02-15
[CB-41] Loaded Objects Strategy

EPecherkin:
[CB-41](https://youtrack.toptal.net/youtrack/issue/CB-41)


features/fixtures/simple_app/spec/spec_helper.rb
```diff
@@ -7,6 +7,7 @@
 require_relative '../../../../lib/crystalball/rails'
 Crystalball::MapGenerator.start! do |c|
   c.register Crystalball::MapGenerator::CoverageStrategy.new
+  c.register Crystalball::MapGenerator::LoadedObjectsStrategy.new
```
EPecherkin: Use `AllocatedObjectsStrategy`

features/change_files_spec.rb
```diff
@@ -22,6 +22,7 @@ class Class1
                                  ./spec/class1_spec.rb[1:1:3:1]
```
EPecherkin: rework it to expect really necessary examples. all over examples should be ignored (sub array matching)

lib/crystalball/map_generator/loaded_objects_strategy/execution_detector.rb
```diff
@@ -0,0 +1,43 @@
...
+            wrapped.source_file&.start_with?(root_path)
+          end.compact
+
+          paths = project_classes.flat_map do |wrapped|
+            wrapped.candidates.flat_map do |candidate|
+              candidate.send(:first_method_source_location).first
+            end
+          end.uniq
+
+          paths
```
EPecherkin: common code for all detectors. move to concerns

lib/crystalball/execution_map.rb
```diff
@@ -37,7 +37,7 @@ def initialize(metadata: {}, cases: {})
     #
     # @param [Crystalball::CaseMap]
     def <<(case_map)
-      cases[case_map.uid] = case_map.affected_files
+      cases[case_map.uid] = case_map.affected_files.uniq
```
pluff, line 40: maybe we can just use set in `CaseMap#affected_files`?

lib/crystalball/map_generator/allocated_objects_strategy/definition_tracer.rb
```diff
@@ -0,0 +1,34 @@
...
+        attr_reader :trace_point, :constants_definition_paths
+
+        def initialize
+          @constants_definition_paths = {}
+        end
+
+        def start
+          self.trace_point ||= TracePoint.new(:class) do |tp|
+            class_name = tp.binding.eval('name')
+            next unless class_name && tp.path
```
pluff, line 18: when it happens?

lib/crystalball/map_generator/allocated_objects_strategy/definition_tracer.rb
```diff
@@ -0,0 +1,34 @@
...
+            constants_definition_paths[class_name] = tp.path
```
pluff: I believe a class can have multiple definition paths

lib/crystalball/map_generator/allocated_objects_strategy/object_lister.rb
```diff
@@ -0,0 +1,61 @@
...
+        # @return Set[Integer] object_id list of allocated objects except ignored
+        def list_ids
+          objects = Set.new
+
+          ObjectSpace.each_object(Object) do |object|
+            next if ignored_types.include?(object.class)
+            objects << object.__id__
+          end
+
+          objects
```
pluff: what do you think about "marking" objects instead of saving them to a list.
Something like this:
```ruby
         ObjectSpace.each_object(Object) do |object|
            next if ignored_types.include?(object.class)
            object.instance_variable_set(:@crystalball_old_object, true)
          end
#.... rspec example here ....
          ObjectSpace.each_object(Object) do |object|
            next if ignored_types.include?(object.class) || object.instance_variable_get(:@crystalball_old_object)
            objects << object
          end
```
Then you'll be able to remove search logic since every object already has info if he's old or new for us

lib/crystalball/map_generator/execution_detector.rb
```diff
@@ -0,0 +1,25 @@
...
+      # @param[Array<String>] list of paths to process
+      # @return [Array<String>]
+      def detect(paths)
+        paths
+          .select { |file_name| file_name.start_with?(root_path) }
+          .map { |file_name| file_name.sub("#{root_path}/", '') }
+      end
+    end
+  end
+end
```
pluff: This is very poor naming. It doesn't detect anything. It's not useful on it own. See comment in strategy file

lib/crystalball/map_generator/allocated_objects_strategy.rb
```diff
@@ -0,0 +1,67 @@
...
+      def call(case_map)
+        GC.start
+        GC.disable
+
+        objects = object_lister.created_during do
+          yield case_map
+        end
+
+        paths = fetch_paths_for_objects(objects)
+        case_map.push(*execution_detector.detect(paths))
```
pluff: In your particular case `fetch_paths_for_objects` is exactly what detector should do. So your `ExecutionDetector` should have `DefinitionTracer` and `HierarchyLister` dependencies while `AllocatedObjectsStrategy` should have `ExecutionDetector` and `ObjectLister` dependencies only

lib/crystalball/map_generator/allocated_objects_strategy/object_lister.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+module Crystalball
+  class MapGenerator
+    class AllocatedObjectsStrategy
+      # Class to list objects from ObjectSpace
+      class ObjectLister
```
pluff: poor name. Consider `ObjectsTracker` or something. `Lister` has completely different meaning

lib/crystalball/map_generator/allocated_objects_strategy/hierarchy_lister.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Crystalball
+  class MapGenerator
+    class AllocatedObjectsStrategy
+      # Class to get full hierarchy of a class(including singleton_class)
+      class HierarchyLister
```
pluff: same problem

lib/crystalball/map_generator/coverage_strategy/execution_detector.rb
```diff
@@ -1,26 +1,20 @@
...
       # Class for detecting code execution path based on coverage information diff
-      class ExecutionDetector
-        attr_reader :root_path
-
-        # @param [String] absolute path to root folder of repository
-        def initialize(root_path)
-          @root_path = root_path
-        end
-
+      class ExecutionDetector < ::Crystalball::MapGenerator::ExecutionDetector
```
pluff: I don't think you need inheritance here. Simple module will do the work

pluff (changes requested):

lib/crystalball/execution_map.rb
EPecherkin: that is a good idea

EPecherkin: or not. We will need to transform affected_files to an array for `YAML`. And change some delegations. I don't see it more readable or fast here

lib/crystalball/map_generator/allocated_objects_strategy/definition_tracer.rb
EPecherkin: discussed

lib/crystalball/map_generator/allocated_objects_strategy/object_lister.rb
EPecherkin: It looks like it works faster. But it is impossible to set instance variable for a frozen object. It shouldn't be a lot of them, but anyway. Maybe we can mix two strategies together

EPecherkin:
rename branch and create new pr

PR #42, opened 2018-02-22
[CB-53] Allocated Objects Strategy

EPecherkin:
[CB-53](https://youtrack.toptal.net/youtrack/issue/CB-53)


features/change_files_spec.rb
```diff
@@ -16,29 +16,48 @@ class Class1
...
+    )
+  end
+
+  it 'generates map if Class1 reopen is changed' do
+    class1_reopen_path.open('w') { |f| f.write <<~RUBY }
+      class Class1
+      end
+    RUBY
+
+    is_expected.to include(
```
jaimerson, line 36: Why is it `include` now instead of `match_array`? Shouldn't it remain strict?

jaimerson (approved):
lgtm

features/change_files_spec.rb
EPecherkin: I don't think so. It is not that bad if we predict a bit more. So, I think it is better to specify examples which should contain than to specify all output of predictor

jaimerson: Maybe, but these specs allow it to predict a lot more. Is the full output inconveniently longer?

pluff (approved):

pluff:
Please squash everything into meaningful commits

PR #44, opened 2018-03-06
[CB-61] Add versioning for execution map

EPecherkin:


lib/crystalball/execution_map.rb
```diff
@@ -30,7 +36,19 @@ def to_h
...
+      version = metadata[:version].to_f
+
+      if cases.any?
+        version = 1.0 if version.zero?
+        guard_version_compatibility(version)
+      elsif version.nonzero?
+        guard_version_compatibility(version)
+      else
+        version = VERSION
+      end
```
pluff: Map version heavily depends on configuration and particular use, so we shouldn't specify it. It should be specified in outer world instead via Configuration object same as `commit`.
E.g. 
```
MapGenerator.start! do |config|
  config.version = 1
end
```

P.S. Don't forget to check version consistency on map load

pluff (changes requested):

lib/crystalball/execution_map.rb
```diff
@@ -30,7 +36,19 @@ def to_h
     # @param [Hash] initial list of cases
     def initialize(metadata: {}, cases: {})
       @cases = cases
-      @metadata = Metadata.new(type: self.class.name, **metadata)
+
+      version = metadata[:version].to_f
+
+      if cases.any?
+        version = 1.0 if version.zero?
```
jaimerson: This does exactly what `guard_version_compatibility` does.

jaimerson: So this could be done after [this](https://github.com/toptal/crystalball/pull/44/files#diff-5fc912a5f464543e946a3414e981f7f7R40) and avoid repeating

lib/crystalball/execution_map.rb
EPecherkin: It is checked now because https://github.com/toptal/crystalball/blob/master/lib/crystalball/map_storage/yaml_storage.rb#L39 checks the consistency of all metadata fields

jaimerson (approved):

pluff (approved):

PR #47, opened 2018-03-07
Improve documentation for map generation and prediction strategies

jaimerson:
Also, fix some other documentation and add yard check on CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,11 @@ Or install it yourself as:
 ## Usage
 
 1. Start MapGenerator in your `spec_helper` before you loaded any file of your app. E.g.
-```ruby
-Crystalball::MapGenerator.start! do |config|
-  config.register Crystalball::MapGenerator::CoverageStrategy.new
-end
-```
+  ```ruby
+  Crystalball::MapGenerator.start! do |config|
+    config.register Crystalball::MapGenerator::CoverageStrategy.new
+  end
+  ```
 1. Run your test suite on clean branch with green build. This step will create file `execution_map.yml` in your project root
 1. Make some changes to your app code
 1. To see list of tests which might fail because of your changes, call:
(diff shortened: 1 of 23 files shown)


pluff:
I believe we should add information about our strategies to README.md too

README.md
```diff
@@ -42,6 +42,25 @@ Crystalball.foresee do |predictor|
 end
 ```
 
+## Map Generator
+
+There are different map generator strategies that can (and should) be used together for better predictions. Each one has its own benefits and drawbacks, so they should be configured to best fit your needs.
+
+- **CoverageStrategy** uses coverage information to detect which files are covered by the given spec (i.e. the files that, if changed, may potentially break the spec);
+- **AllocatedObjectsStrategy** looks for the files in which the objects created during the spec execution are defined. It is considerably slower than `CoverageStrategy`.
```
pluff: I believe we need to add more information on how to configure & extend strategies. 
Please mention DI we have and what they are used for, also mention convenience `build` method for `AllocatedObjectsStrategy`. Fully configured example will be useful
Also we need information about Rails views strategy.

lib/crystalball/case_map.rb
```diff
@@ -8,8 +8,8 @@ class CaseMap
 
     delegate %i[push] => :affected_files
 
-    # @param [String] id of example
-    # @param [Array<String>] list of files affected by example
+    # @param [String] example - id of example
```
pluff, line 11: its not an id its an example itself

pluff (changes requested):

README.md
```diff
@@ -42,6 +42,25 @@ Crystalball.foresee do |predictor|
...
+
+## Predictor
+
+The predictor can also be customized with different strategies:
+
+- **AssociatedSpecs** needs to be configured with rules for detecting which specs should be on the prediction.
+  `predictor.use Crystalball::Predictor::AssociatedSpecs.new(from: %r{models/(.*).rb}, to: "./spec/models/%s_spec.rb")`
```
jaimerson: Change it as per https://github.com/toptal/crystalball/pull/50

EPecherkin: and following https://github.com/toptal/crystalball/pull/51

README.md
```diff
@@ -42,6 +42,25 @@ Crystalball.foresee do |predictor|
...
+  will add `./spec/models/foo_spec.rb` to prediction when `models/foo.rb` changes.
+  This strategy does not depend on a previoulsy generated case map.
+- **ModifiedExecutionPaths** checks the case map and the diff to see which specs are affected by the new or modified files.
+- **ModifiedSpecs**, as the name implies, checks for modified specs. The scope can be modified by passing a regex as argument, which defaults to `%r{\Aspec/.*_spec\.rb\z}`.
```
EPecherkin: the new default regex is going to be `%r{spec/.*_spec\.rb\z}` 
check https://github.com/toptal/crystalball/pull/51/files

README.md
EPecherkin: we definitely need to add more info for all strategies

lib/crystalball/case_map.rb
jaimerson: Then it was wrong before, and also shouldn't be a `String`

README.md
jaimerson: @EPecherkin can you be more specific?

lib/crystalball/case_map.rb
pluff: It was wrong, but it is a string. E.g.`my_spec.rb[1:2:1]`

pluff (approved):

README.md
```diff
@@ -42,6 +42,117 @@ Crystalball.foresee do |predictor|
...
+This strategy will take each example that has a `described_class` (i.e. examples inside `describe` blocks of classes and not strings) and add the paths where the described class and its ancestors are defined to the case map of the example;
+
+To use it, add to your `Crystalball::MapGenerator.start!` block:
+
+```ruby
+# ...
+config.register Crystalball::MapGenerator::DescribedClassStrategy.new
+```
+
+As with `CoverageStrategy`, you can pass a custom execution detector (an object that responds to `#detect` and returns the paths) to the initialization:
```
EPecherkin: Execution detector for DescribedClassStrategy is more like for AllocatedObjectsStrategy

README.md
```diff
@@ -42,6 +42,117 @@ Crystalball.foresee do |predictor|
...
+To use it, add to your config:
+
+```ruby
+# ...
+config.register Crystalball::MapGenerator::I18nStrategy.new
+```
+
+### Custom strategies
+
+You can create your own strategy and use it with the map generator. Any object that responds to `#call` receiving a `Crystalball::CaseMap` and augmenting its list of affected files using `case_map.push(*paths_to_files)`.
```
EPecherkin: `call` receives not a case_map only. It receives example as well

EPecherkin (approved):

PR #49, opened 2018-03-07
[CB-68] Add described_class to execution map

EPecherkin:


lib/crystalball/case_map.rb
```diff
@@ -3,17 +3,21 @@
...
 
     delegate %i[push] => :affected_files
+    delegate %i[file_path] => :example
 
     # @param [String] id of example
     # @param [Array<String>] list of files affected by example
     def initialize(example, affected_files = [])
-      @uid = example.id
-      @file_path = example.file_path
+      @example = example
```
pluff: No!
It saves the reference of an example => never GC.

lib/crystalball/map_generator/allocated_objects_strategy.rb
```diff
@@ -35,6 +35,8 @@ def call(case_map)
         objects = object_tracker.created_during do
           yield case_map
         end
+        described_class = case_map.example.metadata[:described_class]
+        objects.push(case_map.example.metadata[:described_class]) if described_class.is_a?(Class)
```
pluff: It conflicts with strategy purpose. object wasn't allocated during an example. Why do you this?

pluff (changes requested):

lib/crystalball/map_generator/allocated_objects_strategy.rb
EPecherkin: ok. let's rename strategy to `UsedObjectsStrategy`, wdyt?
because there is a case when map is empty for an example
you can check it by yourself. just run rspec for `./spec/class2_spec.rb[1:4:1]` and check the execution_map

lib/crystalball/case_map.rb
EPecherkin: good point. I can store metadata of an example only. wdyt?

lib/crystalball/map_generator/allocated_objects_strategy.rb
pluff: That test is senseless. Are you testing Ruby there?

In theory there might be a case when you call pure class level DSL inside of an extended module, however pushing `described_class` doesn't save you from this flaw since it's just a workaround for one specific case (when your extended class is described_class). 
`UsedObjectsStrategy` is also poor name because it doesn't return all used objects. E.g.
```
class A; end
class B; end
describe A do
 specify { expect(described_class.ancestors.include?(B)).to eq false }
end
```
B is used but won't be in a map. So I prefer a strategy to do fully what it claims to do rather than a strategy doing a bit more with side effects.

pluff: You can consider creating a separate strategy a replacement for prediction strategy `AssociatedSpecs.new from: 'model.rb', to: 'model_spec.rb'`.

Something that reuses the same DefinitionTracer and adds `described_class` definition_paths to the map. That will cover slightly more than current prediction strategy because it will handle cases where a class spec is distributed across several files\contexts.

lib/crystalball/case_map.rb
pluff: Let's keep simple things simple. two strings are way better than one unknown object.

lib/crystalball/map_generator/described_class_strategy.rb
```diff
@@ -0,0 +1,38 @@
...
+      attr_reader :execution_detector
+
+      delegate %i[after_register before_finalize] => :execution_detector
+
+      def self.build(only: [], root: Dir.pwd)
+        hierarchy_fetcher = ClassSourcesDetector::HierarchyFetcher.new(only)
+        execution_detector = ClassSourcesDetector.new(root_path: root, hierarchy_fetcher: hierarchy_fetcher)
+
+        new(execution_detector: execution_detector)
+      end
```
pluff: do we really need `only` part?

lib/crystalball/map_generator/described_class_strategy.rb
```diff
@@ -0,0 +1,38 @@
...
+      def initialize(execution_detector:)
+        @execution_detector = execution_detector
+      end
+
+      def call(case_map, example)
+        yield case_map
+
+        described_class = example.metadata[:described_class]
+
+        case_map.push(*execution_detector.detect([described_class])) if described_class
```
pluff, line 27: described_class is nil for string describes? E.g. `describe "mymodel" {}`

lib/crystalball/map_generator/class_sources_detector.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require 'crystalball/map_generator/helpers/path_filter'
+require 'crystalball/map_generator/class_sources_detector/hierarchy_fetcher'
+require 'crystalball/map_generator/class_sources_detector/definition_tracer'
+
+module Crystalball
+  class MapGenerator
+    # Class for files paths affected object definition
+    class ClassSourcesDetector
```
pluff: I guess it should be `ObjectSourcesDetector` since it works with singleton classes too

lib/crystalball/map_generator/described_class_strategy.rb
EPecherkin: Decided to keep it

features/fixtures/simple_app/spec/spec_helper.rb
```diff
@@ -7,7 +7,9 @@
 require_relative '../../../../lib/crystalball/rails'
 Crystalball::MapGenerator.start! do |c|
   c.register Crystalball::MapGenerator::CoverageStrategy.new
-  c.register Crystalball::MapGenerator::AllocatedObjectsStrategy.build(only: ['Object'])
+  allocated_strategy = Crystalball::MapGenerator::AllocatedObjectsStrategy.build(only: ['Object'])
```
jaimerson: Why are you creating this variable? It does not make it clearer

lib/crystalball/map_generator/described_class_strategy.rb
EPecherkin: yep, it will be nil

features/fixtures/simple_app/spec/spec_helper.rb
EPecherkin: yeah. no need for this variable at all

jaimerson (approved):

jaimerson (approved):

pluff (approved):

PR #54, opened 2018-03-20
Rework Allocated Objects strategy to use TracePoint instead of Object…

pluff:
…Space.

The new approach should be ~2x faster

### DIFF
--- a/spec/map_generator/allocated_objects_strategy/object_tracker_spec.rb
+++ b/spec/map_generator/allocated_objects_strategy/object_tracker_spec.rb
@@ -3,32 +3,42 @@
 require 'spec_helper'
 
 describe Crystalball::MapGenerator::AllocatedObjectsStrategy::ObjectTracker do
-  subject(:tracker) { described_class.new(only_of: only_of) }
-
-  let(:only_of) { ['Object'] }
-  let(:obj1) { double }
-  let(:obj2) { double(class: Dummy) }
+  subject(:tracker) { described_class.new }
 
   before do
     stub_const('Dummy', Class.new)
-    allow(ObjectSpace).to receive(:each_object).with(Object).and_yield(obj1)
+    stub_const('SubDummy', Class.new(Dummy))
   end
 
-  describe '#created_during' do
-    subject do
-      tracker.created_during do
-        allow(ObjectSpace).to receive(:each_object).with(Object).and_yield(obj1).and_yield(obj2)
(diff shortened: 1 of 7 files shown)


spec/map_generator/allocated_objects_strategy/object_tracker_spec.rb
```diff
@@ -3,32 +3,42 @@
...
+    it 'returns objects created during a block' do
+      expect(tracker.used_classes_during do
+        Dummy.new
+        Object.new
+      end).to match_array([Object, Dummy])
     end
 
-    it 'lists objects' do
-      expect(subject).to eq [obj2]
+    it 'otherwise' do
```
EPecherkin: documented output will be
```
Crystalball::MapGenerator::AllocatedObjectsStrategy::ObjectTracker
  #used_classes_during
    otherwise
```
That is not clear when you see it
Name it better please

EPecherkin (approved):
Just a minor change

spec/map_generator/allocated_objects_strategy/object_tracker_spec.rb
```diff
@@ -3,32 +3,42 @@
...
+
+      it 'ignores created objects of other classes' do
+        expect(tracker.used_classes_during do
+          Dummy.new
+          Object.new
+        end).to match_array(Dummy)
+      end
+
+      it 'lists subclasses too' do
+        expect(tracker.used_classes_during { SubDummy.new }).to match_array(SubDummy)
```
jaimerson, line 40: This should be `expect(tracker.used_classes_during { Dummy.new }).to match_array(SubDummy)` or it does not test what the description says

spec/map_generator/allocated_objects_strategy/object_tracker_spec.rb
jaimerson: :+1:
`it 'returns empty array when no object is allocated'` for example

jaimerson (changes requested):

spec/map_generator/allocated_objects_strategy/object_tracker_spec.rb
pluff: `with only_of specified it lists subclasses too`. I'm not sure what you feel wrong about description and it does exactly what it should do. It lists subclasses of specified `only_of`

jaimerson (approved):

PR #56, opened 2018-03-22
Create GraphQL map generation strategy

jaimerson:


lib/crystalball/graphql/map_generator/graphql_strategy/patch.rb
```diff
@@ -0,0 +1,45 @@
...
+                end
+              end
+            end
+          end
+
+          # Will replace original `GraphQL::ObjectType#define`.
+          # Pushes path of files where `ObjectType`s are defined.
+          # and then calls the original `define`
+          def cb_patched_define(*args, **kwargs, &block)
+            GraphQLStrategy.type_definition_paths.concat(caller_locations.map(&:path))
```
pluff: It will return huge list. Why do you need it?

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,42 @@
...
+      class GraphQLStrategy
+        include ::Crystalball::MapGenerator::BaseStrategy
+        include ::Crystalball::MapGenerator::Helpers::PathFilter
+
+        class << self
+          attr_accessor :type_definition_paths
+        end
+
+        def after_register
+          GraphQLStrategy.type_definition_paths = []
```
pluff: you can use self.class

lib/crystalball/graphql/map_generator/graphql_strategy/schema_execution_detector.rb
```diff
@@ -0,0 +1,40 @@
...
+            # @return [Boolean] whether schema was executed
+            def schema_executed?
+              self.executed = false
+              old_strategy = ::GraphQL::Schema.default_execution_strategy
+              ::GraphQL::Schema.default_execution_strategy = Strategy
+              yield
+              executed
+            ensure
+              ::GraphQL::Schema.default_execution_strategy = old_strategy
+            end
```
pluff: I'm sure it won't work in 100% cases

lib/crystalball/graphql/map_generator/graphql_strategy/schema_execution_detector.rb
```diff
@@ -0,0 +1,40 @@
...
+
+            # Simple GraphQL execution strategy that marks as executed and
+            # forwards to default implementation
+            class Strategy < ::GraphQL::Execution::Execute
+              def execute(*args)
+                SchemaExecutionDetector.executed = true
+                super(*args)
```
pluff: Let's discuss it. I think GraphQL load types not only when you execute a query.

pluff (changes requested):

lib/crystalball/graphql/map_generator/graphql_strategy/schema_execution_detector.rb
jaimerson: Well, that's what [they say](https://github.com/rmosolgo/graphql-ruby/blob/952bb07367a810dbedf930b5133a19aba9ffd2f3/lib/graphql/define/instance_definable.rb#L12), and you can confirm it by breaking a `define` block, running a spec that does not execute `GraphQL` queries and seeing it pass.

lib/crystalball/graphql/map_generator/graphql_strategy/schema_execution_detector.rb
jaimerson: Why?

lib/crystalball/graphql/map_generator/graphql_strategy/patch.rb
jaimerson: Indeed. It's being filtered in `GraphQLStrategy#call` but we still keep it in the class instance variable. Do you know a better way to get the definition paths? Or maybe it's better to just filter it here

lib/crystalball/graphql/map_generator/graphql_strategy/schema_execution_detector.rb
jaimerson: Change it to `ensure_defined`

EPecherkin (commented):
I have nothing special to add to Pavel's comments. Just fix them and I will check it again and approve

lib/crystalball/graphql/map_generator/graphql_strategy/instance_definable_patch.rb
```diff
@@ -0,0 +1,31 @@
...
+              end
+            end
+          end
+
+          # Will replace original `GraphQL::Define::InstanceDefinable.ensure_defined`.
+          # Marks `executed` as true and then calls the original `define`
+          def ensure_defined(*)
+            ::GraphQL::Define::InstanceDefinable.executed = true
+            super
+          end
```
pluff: why do we have to pollute GraphQL module with our flag accessor?
where is revert functionality?

pluff: It also doesn't save the type that was defined. There might be multiple schemas defined and they'll have separate definition trees.

lib/crystalball/graphql/map_generator/graphql_strategy/object_type_patch.rb
```diff
@@ -0,0 +1,45 @@
...
+                end
+              end
+            end
+          end
+
+          # Will replace original `GraphQL::ObjectType#define`.
+          # Pushes path of files where `ObjectType`s are defined.
+          # and then calls the original `define`
+          def cb_patched_define(*args, **kwargs, &block)
+            GraphQLStrategy.type_definition_paths.concat(caller_locations.map(&:path))
```
pluff: can you send me an example of what will be in definition paths?

pluff: same question. there might be multiple different schemas in the project

pluff (changes requested):

lib/crystalball/graphql/map_generator/graphql_strategy/object_type_patch.rb
jaimerson: https://gist.github.com/jaimerson/cd9786c2daff3bb827751715fe522dc5

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,71 @@
...
+          def initialize
+            super
+            @tracers = [Tracer]
+          end
+        end
+
+        def initialize(root: Dir.pwd)
+          @type_definition_paths = []
+          @root_path = root
+          @trace = TracePoint.new(:call) do |tp|
```
pluff: what is peformance overhead here? Please measure

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,71 @@
...
+            GraphQLStrategy.executed = true
+            yield
+          end
+        end
+
+        # Adds the `GraphQLStrategy::Tracer` to all `GraphQL::Schema`s
+        module Patch
+          def initialize
+            super
+            @tracers = [Tracer]
```
pluff, line 79: I dont see where it is used

jaimerson: It's internal to `GraphQL::Schema`. http://graphql-ruby.org/queries/tracing.html

features/fixtures/simple_app/spec/graphql_schema1_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+describe Schema1 do
+  describe '.types' do
+    subject(:types) { described_class.types }
+
+    it 'includes Type1' do
+      expect(types).to include('Type1')
+    end
+
+    it 'executes queries' do
+      described_class.execute('{foo}')
```
EPecherkin, line 14: How does it work? Do you expect here something? Or you just run it to check no exception raised?

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,125 @@
...
+
+          case_map.push(*paths)
+        end
+
+        private
+
+        def tracer
+          @tracer ||= TracePoint.new(:call) do |tp|
+            next unless tp.method_id == :ensure_defined
+            this = tp.binding.eval('self')
```
EPecherkin, line 114: `tp.self`?

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,125 @@
...
+          self.class.type_definition_paths = {}
+          @root_path = root
+        end
+
+        def after_register
+          self.class.apply_patch
+          tracer.enable
+        end
+
+        def before_finalize
```
EPecherkin, line 93: Don't we need to revert Patch here?

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,125 @@
...
+          end
+
+          def type_definition_paths_during
+            apply_patch! unless patch_applied
+            self.current_schema = nil
+
+            yield
+
+            return [] unless current_schema
+            type_definition_paths[current_schema.object_id]
```
pluff, line 55: `type_definition_paths[current_schema.object_id] || []`

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,125 @@
...
+          end
+        end
+
+        attr_accessor :root
+
+        # Tracer to run on every GraphQL query.
```
pluff, line 61: will it work for cases where we don't execute a query? E.g. when we just load the type of a schema

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,125 @@
...
+          tracer.disable
+        end
+
+        # Adds to the map the files in which `GraphQL::ObjectType.define` is called.
+        # Filters paths to only those inside root path.
+        # Raises error if `::GraphQL` is not defined.
+        # @param [Crystalball::CaseMap] case_map - object holding example metadata and affected files
+        def call(case_map, _)
+          paths = self.class.type_definition_paths_during do
+            yield case_map
```
pluff, line 103: pass all other arguments to yield too. Your `call` doesn't use it, however child `calls` could

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,125 @@
...
+
+        def tracer
+          @tracer ||= TracePoint.new(:call) do |tp|
+            next unless tp.method_id == :ensure_defined
+            this = tp.binding.eval('self')
+
+            paths = filter(caller_locations.map(&:path).reject { |p| p =~ %r{/spec} })
+
+            self.class.type_definition_paths[this.object_id] ||= []
+            self.class.type_definition_paths[this.object_id]  |= paths
```
pluff, line 119: maybe create method `add_definition_paths(this, paths)`?

lib/crystalball/graphql/map_generator/graphql_strategy.rb
```diff
@@ -0,0 +1,125 @@
...
+          def augment_type_definition_paths(key, other_keys)
+            type_definition_paths[key.object_id] |= other_keys
+                                                    .map { |t| type_definition_paths[t.object_id] }
+                                                    .flatten
+                                                    .compact
+          end
+
+          # Checks whether `::GraphQL::Schema` is defined and, if it is, applies
+          # the patch to it. It needs to be this way because `graphql` gem is weird.
+          def apply_patch
```
pluff, line 29: why do you need this method at all? Why would you want to include new strategy which is senseless if you don't have GraphQL?

pluff (changes requested):

features/fixtures/simple_app/spec/graphql_schema1_spec.rb
jaimerson: It does not need to do anything really, not even be a valid query, just execute it

lib/crystalball/graphql/map_generator/graphql_strategy.rb
jaimerson: No, but there's no point, or at least I haven't thought of any situations where you can use the type in any other way. If the example does not execute any queries, you're probably testing graphql types in a weird way and it will be caught by associated_specs strategy when executing.

lib/crystalball/graphql/map_generator/graphql_strategy.rb
jaimerson: GraphQL specs break when we require it in this file. I'll just keep the bang version to be called during `#call` where we're sure `graphql` has been loaded already.

PR #59, opened 2018-04-03
Add RSpec runner for Crystalball predictions

pluff:
This should improve convenience level for local predictions significantly.


bin/crystalball
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
```
jaimerson, line 2: I wonder if it does anything not being the first line

lib/crystalball/rspec/runner.rb
```diff
@@ -0,0 +1,38 @@
...
+    # Our custom RSpec runner to run predictions
+    class Runner < ::RSpec::Core::Runner
+      class << self
+        def invoke(config)
+          setup_prediction_builder(config)
+          super()
+        end
+
+        def run(args, err = $stderr, out = $stdout)
+          out.puts "Crystalball starts to glow..."
```
jaimerson, line 17: Haha love it

lib/crystalball/rspec/runner.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        attr_reader :prediction_builder
+
+        def setup_prediction_builder(config)
+          @prediction_builder = PredictionBuilder.new(config)
+        end
+
+        def build_prediction(out)
+          prediction = prediction_builder.prediction.compact
+          out.puts "Prediction: #{prediction.first(5).join(' ')}#{'...' if prediction.size > 5}"
```
jaimerson, line 31: Can we make this configurable via the yml?

jaimerson (approved):
Looks good, one minor comment

bin/crystalball
pluff: it does

lib/crystalball/rspec/runner.rb
pluff: I personally don't see value in this configuration. It doesn't affect the library behavior. It just truncates huge output. Usually, you care about first items only since it can be something like `spec/` or `./`. For large projects loading entire suite can take a while and I wanted to give a user an opportunity to react earlier.

PR #65, opened 2018-04-26
[CB-116] Add ModifiedSupportSpecs predictor

EPecherkin:


lib/crystalball/predictor/modified_support_specs.rb
```diff
@@ -0,0 +1,35 @@
...
+
+module Crystalball
+  class Predictor
+    # Used with `predictor.use Crystalball::Predictor::ModifiedSupportSpecs.new`. Will find files that match passed regexp and
+    # return full spec files which uses matched support spec files. Perfectly works for shared_context and shared_examples.
+    class ModifiedSupportSpecs
+      include ::Crystalball::Predictor::Helpers::AffectedExamplesDetector
+
+      # @param [Regexp] support_spec_pattern - regexp to filter support specs files
+      def initialize(support_spec_pattern = %r{spec/support/.*\.rb\z})
```
pyromaniac, line 13: Should not this default be somewhere in settings?

pyromaniac (approved):
LG if it works

jaimerson (approved):
LGTM

lib/crystalball/predictor/modified_support_specs.rb
EPecherkin: It is a common way to configure predictors right now. I think it makes sense to move it to config file later

PR #66, opened 2018-04-30
Create parser strategy

jaimerson:
This strategy is meant to detect calls to class methods and add the referenced class to the case map.

### DIFF
--- a/crystalball.gemspec
+++ b/crystalball.gemspec
@@ -31,6 +31,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency 'git'
+  spec.add_dependency 'parser'
 
   spec.required_ruby_version = '> 2.3.0'
 
--- a/crystalball.gemspec
+++ b/crystalball.gemspec
@@ -31,13 +31,13 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency 'git'
-  spec.add_dependency 'parser'
 
   spec.required_ruby_version = '> 2.3.0'
 
   spec.add_development_dependency 'actionview'
   spec.add_development_dependency "bundler", "~> 1.14"
   spec.add_development_dependency 'i18n'
+  spec.add_development_dependency 'parser'
   spec.add_development_dependency 'pry'
   spec.add_development_dependency 'pry-byebug'
   spec.add_development_dependency "rake", "~> 10.0"
--- a/features/modified_execution_paths/class_method_call_spec.rb
+++ b/features/modified_execution_paths/class_method_call_spec.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+require_relative '../feature_helper'
+
+describe 'Changing source file' do
+  include_context 'base forecast'
+  include_context 'simple git repository'
+
+  let(:strategies) { [Crystalball::Predictor::ModifiedExecutionPaths.new] }
+
+  it 'adds class2 example when class1 changes because of class method call' do
+    change class1_path
+
+    expect(forecast).to include(
+      './spec/class2_spec.rb[1:4:1]'
+    )
+  end
+end
--- a/lib/crystalball/map_generator/parser_strategy.rb
+++ b/lib/crystalball/map_generator/parser_strategy.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'crystalball/map_generator/parser_strategy/processor'
+require 'crystalball/map_generator/helpers/path_filter'
+
+module Crystalball
+  class MapGenerator
+    # Map generator strategy based on parsing source files to detect constant definition
+    # and tracing method calls on those constants.
+    class ParserStrategy
+      include BaseStrategy
+      include Helpers::PathFilter
+
+      def initialize(root = Dir.pwd, pattern:)
+        @root_path = Pathname(root).realpath.to_s
+        @processor = Processor.new
+        @const_definition_paths = {}
+        @pattern = pattern
+      end
+
+      def after_register
+        files_to_inspect.each do |path|
+          processor.consts_defined_in(path).each do |const|
+            (const_definition_paths[const] ||= []) << path
+          end
+        end
+      end
+
+      # Parses the current case map seeking calls to class methods and adds
+      #   the classes to the map.
+      # @param [Crystalball::CaseMap] case_map - object holding example metadata and affected files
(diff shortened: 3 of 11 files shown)


features/modified_execution_paths/class_method_call_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+describe 'Changing source file' do
+  subject(:forecast) do
+    Crystalball.foresee(workdir: root, map_path: root.join('execution_map.yml')) do |predictor|
+      predictor.use Crystalball::Predictor::ModifiedExecutionPaths.new
+    end
+  end
+  include_context 'simple git repository'
+
+  it 'adds class2 example when class1 changes because of class method call' do
+    change class1_path
```
EPecherkin: This is not enough to test it. This case will be handled as well by coverage strategy

lib/crystalball/map_generator/parser_strategy/processor.rb
```diff
@@ -0,0 +1,123 @@
...
+        end
+
+        def on_casgn(node, scope)
+          namespace, name, = node.children
+          scope_name = namespace ? qualified_name(qualified_name_from_node(namespace), scope) : scope
+          add_constant_defined(name, scope_name)
+        end
+
+        def on_class(node, scope)
+          const, _, body = node.children # TODO: _ is superclass
```
EPecherkin: why `TODO` ? are you going to handle it in some way?

jaimerson: Yes, I'll add it to the map as well

EPecherkin: write a more detailed description in such case. Like what should be done here why it isn't

crystalball.gemspec
```diff
@@ -31,6 +31,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency 'git'
+  spec.add_dependency 'parser'
```
pluff: I think this shouldn't be a default dependency. Let's move it to development dependency only.
We need to mention in README that if you want to use parser strategy you need to add `parser` to your Gemfile.

features/modified_execution_paths/class_method_call_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    is_expected.to include(
+      './spec/class2_spec.rb[1:4:1]'
+    )
```
pluff: Please reuse [this](https://github.com/toptal/crystalball/pull/68/files#diff-0fe6e382aea1d1671075e5319b722f6cR7) when it will be merged

lib/crystalball/map_generator/parser_strategy.rb
```diff
@@ -0,0 +1,61 @@
...
+      def files_to_inspect
+        Dir.glob(File.join(root_path, '**/*.rb')).grep(pattern)
+      end
+
+      def tracer_for(case_map)
+        TracePoint.new(:call) do |tp|
+          classes = [tp.defined_class.to_s, tp.defined_class.singleton_class.to_s]
+          used = const_definition_paths.select { |k, _| classes.include?(k) }.keys
+          case_map.push(*used)
+        end
```
pluff: where this method is used?

lib/crystalball/map_generator/parser_strategy/processor.rb
```diff
@@ -0,0 +1,124 @@
...
+          parse_and_process(path)
+          consts_defined
+        end
+
+        def consts_interacted_with_in(path)
+          self.current_scope = nil
+          self.consts_interacted_with = []
+          parse_and_process(path)
+          consts_interacted_with
+        end
```
pluff: why all these handlers are public?

lib/crystalball/map_generator/parser_strategy/processor.rb
```diff
@@ -0,0 +1,124 @@
...
+
+require 'parser/current'
+
+module Crystalball
+  class MapGenerator
+    class ParserStrategy
+      # Parses the given source files and adds the class and module definitions
+      # to the `consts_defined` array.
+      class Processor
+        attr_accessor :consts_defined, :consts_interacted_with, :current_scope
```
pluff: why these are public?

pluff (changes requested):
Did you test the quality of `parser` AST? Does it handle corner cases? E.g. `MyConst = Class.new` and `MyConst = Object.new`?

lib/crystalball/map_generator/parser_strategy.rb
jaimerson: Oh, I forgot to remove it

lib/crystalball/map_generator/parser_strategy/processor.rb
jaimerson: You're right, they shouldn't be. They started like this because this class sublclassed `Parser::AST::Processor`, but that's no longer the case. I'll change it.

jaimerson:
@pluff yes, it handles constant assignment as well https://github.com/toptal/crystalball/blob/ddd2f4fc79550084eb7ce7cc05a7d20267b3efd2/spec/map_generator/parser_strategy/processor_spec.rb#L110-L125
And also `CONSTANT = :foo`

features/modified_execution_paths/class_method_call_spec.rb
EPecherkin: This is still not enough

lib/crystalball.rb
```diff
@@ -14,6 +14,7 @@
 require 'crystalball/map_generator/configuration'
 require 'crystalball/map_generator/coverage_strategy'
 require 'crystalball/map_generator/allocated_objects_strategy'
+require 'crystalball/map_generator/parser_strategy'
```
EPecherkin: I think parser strategy should be included separately in favor of https://github.com/toptal/crystalball/pull/66/files#r185799760

lib/crystalball/map_generator/parser_strategy/processor.rb
EPecherkin: This one still needs a description

features/modified_execution_paths/class_method_call_spec.rb
jaimerson: @EPecherkin I changed `Class1.foo` to be an empty method and confirmed it is not caught by coverage or any of the other strategies.

pluff (approved):

EPecherkin (approved):

PR #67, opened 2018-05-03
Add tables_map_path to crystalball.yml

EPecherkin:
I'll create a PR in another project to show how do I use it


lib/crystalball/predictor.rb
```diff
@@ -3,18 +3,20 @@
 module Crystalball
   # Class to predict test failures with given execution map and sources diff
   class Predictor
-    attr_reader :map, :from, :to, :prediction_strategies
+    attr_reader :map, :from, :to, :prediction_strategies, :configuration
```
pluff: why do we need this? Configuration is for runner, not for base predictor

lib/crystalball/rspec/runner/configuration.rb
```diff
@@ -7,9 +7,10 @@ module RSpec
     class Runner
       # Class for storing local runner configuration
       class Configuration
-        def initialize(config = {})
+        def initialize(config = {}) # rubocop:disable Metrics/MethodLength
           @values = {
-            'map_path' => 'tmp/execution_maps',
+            'execution_map_path' => 'tmp/execution_map.yml',
+            'tables_map_path' => 'tmp/tables_map.yml',
```
pluff: I don't think we need default values here since this tables_map file is optional to one strategy only

spec/fixtures/crystalball.yml
```diff
@@ -1,5 +1,7 @@
-# Custom path to your maps folder. Default: `tmp/execution_maps`
-map_path: 'execution_map.yml'
+# Custom path to your execution map. Can be file or folder. Default: `tmp/execution_map.yml`
+execution_map_path: 'execution_map.yml'
+# Custom path to your tables map. Default: `tmp/tables_map.yml`
```
pluff: Please mention a strategy which uses this value

jaimerson: Also remember to update here if you remove the default

jaimerson (approved):
Looks good, just address Pavel's comments

spec/rspec/runner/configuration_spec.rb
```diff
@@ -27,15 +27,16 @@
...
         'repo_path' => 'test',
         'predictor_class_name' => 'MyPredictor',
         'requires' => ['test.rb'],
         'diff_from' => 'HEAD~3',
         'diff_to' => 'HEAD',
         'map_expiration_period' => 1,
         'runner_class_name' => 'MyRunner',
-        'custom' => 42
+        'custom' => 42,
+        'tables_map_path' => 'tables_map.yml'
```
pluff: no need to add this since `custom` check checks it all. There is no logic related to `tables_map_path` at all.

pluff (dismissed):

features/fixtures/simple_app/crystalball.yml
```diff
@@ -1,5 +1,7 @@
-# Custom path to your maps folder. Default: `tmp/execution_maps`
-map_path: 'execution_map.yml'
+# Custom path to your execution map. Can be file or folder. Default: `tmp/execution_map.yml`
+execution_map_path: 'execution_map.yml'
+# Custom path to your tables map. Used by 'Crystalball::Rails::Predictor::ModifiedSchema'
+tables_map_path: 'tables_map.yml'
```
pluff: where it is used? Why it should be here?

pluff (approved):

lib/crystalball/rspec/prediction_builder.rb
```diff
@@ -39,8 +39,17 @@ def repo
...
+      # def base_predictor
+      #   @base_predictor ||= Crystalball::Predictor.new(execution_map, repo, from: config['diff_from'], to: config['diff_to']) do |p|
+      #     p.use Crystalball::Predictor::ModifiedExecutionPaths.new
+      #     p.use Crystalball::Predictor::ModifiedSpecs.new
+      #   end
+      # end
+      #
       def base_predictor
-        @base_predictor ||= config['predictor_class'].new(map, repo, from: config['diff_from'], to: config['diff_to'])
+        raise NotImplementedError, 'Configure `prediction_builder_class_name` in `crystalball.yml`'
```
jaimerson: Isn't it an Argument Error in this case?

jaimerson (approved):

lib/crystalball/rspec/prediction_builder.rb
EPecherkin: No, it is not implemented. Because users have to implement this method in an ancestor

pluff: Its too easy to break something by dropping part of required params to predictor. This makes Crystalball literally dependable on user implementation. Consider creating `predictor` method as an extension point
E.g.
```
def predictor
  raise NotImplementedError unless block_given?
  @predictor ||= Crystalball::Predictor.new(map, repo, from: config['diff_from'], to: config['diff_to'], &block)
end
```
so you can use it as 
```
def predictor
  super do |p|
    p.use MyStrategy.new(config['my_param'])
  end
end
```

P.S. yeah lets get rid of `base_` prefix

lib/crystalball/rspec/runner/configuration.rb
```diff
@@ -1,18 +1,16 @@
...
       # Class for storing local runner configuration
       class Configuration
         def initialize(config = {})
           @values = {
-            'map_path' => 'tmp/execution_maps',
+            'execution_map_path' => 'tmp/execution_map.yml',
             'map_expiration_period' => 86_400,
             'repo_path' => Dir.pwd,
-            'predictor_class_name' => 'Crystalball::SimplePredictor',
+            'prediction_builder_class_name' => nil,
```
pluff: ~simple predictor~simple prediction builder is actually a good idea to provide something out of the box

lib/crystalball/rspec/runner/configuration.rb
```diff
@@ -35,11 +33,13 @@ def [](key)
 
         private
 
-        def predictor_class
-          @predictor_class ||= begin
+        def prediction_builder_class
+          return nil unless self['prediction_builder_class_name']
```
pluff: It will be `no method "new" on nil` or smth

pluff (changes requested):

lib/crystalball/rails/tables_map_generator.rb
```diff
@@ -67,14 +67,20 @@ def repo
...
-          next if descendant == ActiveRecord::Base || table_name.nil?
+          next if table_name.nil?
 
           map[table_name] = object_sources_detector.detect([descendant])
         end
       end
 
+      def table_name_of(model)
+        model.table_name
+      rescue NoMethodError # Some inner classes catched here
```
pluff: are you sure its better than `model.table_name if model.respond_to?(:table_name)`

EPecherkin: Yeah, too sad but I'm sure

pluff (changes requested):
please add a feature for has_and_belongs_to_many edge case + ensure that a feature for regular model declaration exists

pluff:
Also in future please try to use separate PRs for small changes. E.g. parser strategy fix can be in separate PR, config extension point can be in separate PR, renames can be in separate PR

features/fixtures/simple_app/db/schema.rb
```diff
@@ -12,7 +12,7 @@
     t.index "name"
   end
 
-  create_table "dummies", id: :serial, force: :cascade
+  create_table "model2s", id: :serial, force: :cascade
 
   add_foreign_key "model1s", "dummies", column: "dummy_id"
```
pluff: where is the new habtm table? Where is a feature for it?

pluff (changes requested):

features/modified_schema/change_spec.rb
```diff
@@ -12,8 +12,16 @@
...
 
   it 'adds mapped examples to a prediction list' do
-    change schema_path
+    change schema_path, File.read(fixtures_path.join('schema', 'model_table_changed.rb'))
 
     expect(forecast).to include(*model1_examples)
+    expect(forecast).not_to include('./spec/models/model2_spec.rb[1:1:1]')
+  end
+
+  it 'wirtes a warning for table with no model files' do
```
pluff: typo

README.md
```diff
@@ -175,6 +175,10 @@ Mostly usable for shared_contexts and shared_examples.
 Checks for modified db schema in rails application. You need to specify a path to a file with tables map generated by `TablesMapGenerator`. It checks schema diff to see which models are affected by modified tables and which specs are affected by this models.
 `predictor.use Crystalball::Rails::Predictor::ModifiedSchema.new(tables_map_path: './tables_map.yml')`
 
+#### Warning
+
+You may meet a warning like "WARNING: there are no model files for changing table ...". Usually, such tables are leftovers or a relation table for `has-and-belongs-to-many` associations. For the first case - nothing to worry about. For the second case - it means you want to change the default relation table. We would recommend you to create a model for such table. It may improve prediction for future.
```
pluff: `changed table`

pluff (approved):

PR #69, opened 2018-05-04
Update gem authors

pluff:
We have to fetch repository remotes in case we are loading the map generated on commit which is not present in local working tree (e.g. generated on CI with fresh codebase)


jaimerson (approved):
Please update the description to also describe the changes in repo fetch, so we can easily search for changes in prediction builder later

EPecherkin (approved):

PR #70, opened 2018-05-08
Make prediction builder's repo public

jaimerson:


EPecherkin (approved):

lib/crystalball/rspec/prediction_builder.rb
```diff
@@ -33,12 +33,12 @@ def map
         @map ||= Crystalball::MapStorage::YAMLStorage.load(config['map_path'])
       end
 
-      private
-
       def repo
```
EPecherkin, line 36: add a spec for a public method

PR #78, opened 2018-05-29
Allow parser strategy to be extended

jaimerson:
This PR is intended to allow extension of parser strategy with custom post processors.

### DIFF
--- a/crystalball.gemspec
+++ b/crystalball.gemspec
@@ -43,7 +43,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'pry-byebug'
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.0"
-  spec.add_development_dependency 'rubocop'
+  spec.add_development_dependency 'rubocop', ">= 0.56"
   spec.add_development_dependency 'rubocop-rspec'
   spec.add_development_dependency 'simplecov'
   spec.add_development_dependency 'sqlite3'
--- a/lib/crystalball.rb
+++ b/lib/crystalball.rb
@@ -16,7 +16,6 @@ require 'crystalball/map_generator/configuration'
 require 'crystalball/map_generator/coverage_strategy'
 require 'crystalball/map_generator/allocated_objects_strategy'
 require 'crystalball/map_generator/described_class_strategy'
-require 'crystalball/map_generator/parser_strategy'
 require 'crystalball/map_storage/yaml_storage'
 require 'crystalball/version'
 
--- a/lib/crystalball.rb
+++ b/lib/crystalball.rb
@@ -16,6 +16,7 @@ require 'crystalball/map_generator/configuration'
 require 'crystalball/map_generator/coverage_strategy'
 require 'crystalball/map_generator/allocated_objects_strategy'
 require 'crystalball/map_generator/described_class_strategy'
+require 'crystalball/map_generator/parser_strategy'
 require 'crystalball/map_storage/yaml_storage'
 require 'crystalball/version'
 
--- a/lib/crystalball/map_generator/parser_strategy.rb
+++ b/lib/crystalball/map_generator/parser_strategy.rb
@@ -11,8 +11,10 @@ module Crystalball
       include BaseStrategy
       include Helpers::PathFilter
 
+      attr_reader :const_definition_paths
+
       def initialize(root = Dir.pwd, pattern:)
-        @root_path = Pathname(root).realpath.to_s
+        @root_path = Pathname.new(root).realpath.to_s
         @processor = Processor.new
         @const_definition_paths = {}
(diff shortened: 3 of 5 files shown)


lib/crystalball/map_generator/parser_strategy.rb
```diff
@@ -35,15 +37,24 @@ def call(case_map, *_args)
         case_map.each do |path|
           next unless path.end_with?('.rb')
           consts = processor.consts_interacted_with_in(path)
-          used = const_definition_paths.select { |k, _| consts.include?(k) }.values
-          paths.push(*used.flatten)
+          used = definition_paths_for_constants(consts)
+          process(consts, paths)
+          paths.push(*used)
```
pluff: almost entire block body can be moved to `process` method. Also `process` is a poor name. 
```ruby
def used_files(used_consts)
  const_definition_paths.select { |k, _| Array(used_consts).include?(k) }.values.flatten
end

case_map.each do |path||
  next unless path.end_with?('.rb')
  used_consts = processor.consts_interacted_with_in(path)
  paths.push(*used_files(used_consts).flatten)
end
```
Then you can extends `used_files` if you want
```ruby
def used_files(used_consts)
  super + my_handling
end
```

lib/crystalball/rails/helpers/base_schema_parser.rb
```diff
@@ -19,7 +19,7 @@ def initialize
         private
 
         # Store info about call in hash. First argument of method call used as a key
-        def method_missing(method_name, *args, &block) # rubocop:disable Style/MethodMissingSuper
+        def method_missing(method_name, *args, &block) # rubocop:disable Style/MethodMissing
```
pluff: no need to have it at all. It was disabled

pluff (changes requested):

lib/crystalball.rb
```diff
@@ -16,6 +16,7 @@
 require 'crystalball/map_generator/coverage_strategy'
 require 'crystalball/map_generator/allocated_objects_strategy'
 require 'crystalball/map_generator/described_class_strategy'
+require 'crystalball/map_generator/parser_strategy'
```
jaimerson: Bad idea to require it by default, it would make `parser` a required dependency.

EPecherkin (commented):
I have no special notes here. Just fix Pavel's notes and I will review new code and approve it

lib/crystalball/map_generator/parser_strategy.rb
```diff
@@ -11,8 +11,10 @@ class ParserStrategy
       include BaseStrategy
       include Helpers::PathFilter
 
+      attr_reader :const_definition_paths
+
       def initialize(root = Dir.pwd, pattern:)
-        @root_path = Pathname(root).realpath.to_s
+        @root_path = Pathname.new(root).realpath.to_s
```
EPecherkin, line 17: What is the difference?

lib/crystalball/map_generator/parser_strategy.rb
```diff
@@ -21,6 +23,7 @@ def initialize(root = Dir.pwd, pattern:)
       def after_register
         files_to_inspect.each do |path|
           processor.consts_defined_in(path).each do |const|
+            yield const, path if block_given?
```
EPecherkin: do you still need it?

lib/crystalball/map_generator/parser_strategy.rb
jaimerson: None practical, but it makes stubbing [easier]( https://github.com/toptal/crystalball/blob/30f642db156cdc3e8c872bda6e25ce5c4bee1681/spec/map_generator/parser_strategy_spec.rb#L22).

lib/crystalball/map_generator/parser_strategy.rb
jaimerson: Not right now, I guess I'll just remove it.

crystalball.gemspec
```diff
@@ -43,7 +43,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'pry-byebug'
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.0"
-  spec.add_development_dependency 'rubocop'
+  spec.add_development_dependency 'rubocop', "~> 0.56"
```
pluff: why should we add it? Lets be state-of-the-art always :)

pluff (approved):

EPecherkin (approved):

crystalball.gemspec
jaimerson: I can change it to `>=`, because since we have `MethodMissingSuper` config, which is only available after https://github.com/rubocop-hq/rubocop/pull/5811, there will be warnings and inconsistent behavior on CI/locally

PR #81, opened 2018-06-01
Flag the strategy used to predict files in case map

jaimerson:
This will allow us to select only the specs that were inferred
by individual strategies at prediction time. That way we can choose
between faster predictions with lower precision or thorough ones.

### Example case map
```yaml
"./spec/class1_spec.rb[1:1:1]":
  coverage_strategy:
  - spec/support/shared_examples/module1.rb
  - lib/module1.rb
  - lib/class1.rb
  allocated_objects_strategy:
  - lib/class1.rb
  - lib/class1_reopen.rb
  - lib/module1.rb
  described_class_strategy:
  - lib/class1.rb
  - lib/class1_reopen.rb
  - lib/module1.rb
  parser_strategy:
  - lib/module1.rb
```

### DIFF
--- a/lib/crystalball/predictor/helpers/affected_examples_detector.rb
+++ b/lib/crystalball/predictor/helpers/affected_examples_detector.rb
@@ -11,7 +11,13 @@ module Crystalball
         # @return [Array<String>] list of affected examples
         def detect_examples(files, map)
           map.cases.map do |uid, case_map|
-            uid if files.any? { |file| case_map.include?(file) }
+            uid if files.any? do |file|
+              if case_map.is_a?(Array) # support old version of map. TODO: remove it.
+                case_map.include?(file)
+              else
+                case_map.values.flatten.include?(file)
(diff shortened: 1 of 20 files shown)


lib/crystalball/predictor/helpers/affected_examples_detector.rb
```diff
@@ -11,7 +11,13 @@ module AffectedExamplesDetector
         # @return [Array<String>] list of affected examples
         def detect_examples(files, map)
           map.cases.map do |uid, case_map|
-            uid if files.any? { |file| case_map.include?(file) }
+            uid if files.any? do |file|
+              if case_map.is_a?(Array) # support old version of map. TODO: remove it.
```
EPecherkin, line 15: Hm. I don't think we need to support it. How long it will take to generate a new map after you merge this changes? One hour?

EPecherkin (approved):

PR #84, opened 2018-06-06
Use merge-base for diff

EPecherkin:


lib/crystalball/predictor.rb
```diff
@@ -32,7 +32,12 @@ def prediction
     alias cases prediction
 
     def diff
-      repo.diff(from, to)
+      @diff ||= repo.diff(merge_base, to)
+    end
+
+    def merge_base
+      to_or_head = to ? to : 'HEAD'
+      `git --git-dir=#{repo.repo.path} --work-tree=#{repo.dir.path} merge-base #{from} #{to_or_head} 2>&1`.strip
```
pluff: No!
Please do it on proper level. There should be a method for repo which uses ruby-git lib command stuff etc. merge-base is useful general purpose command which we should try to submit to ruby-git library

pluff (changes requested):

lib/crystalball/predictor.rb
jaimerson: Someone tried that [in 2012](https://github.com/ruby-git/ruby-git/pull/43). It's hard to get anything merged there. But I agree that having it on a higher level of abstraction would benefit our own code base, making it easier to reason about, test, etc. Maybe that PR I linked is a good starting point.

lib/crystalball/predictor.rb
```diff
@@ -32,7 +32,12 @@ def prediction
     alias cases prediction
 
     def diff
-      repo.diff(from, to)
+      @diff ||= repo.diff(merge_base, to)
+    end
+
+    def merge_base
+      to_or_head = to ? to : 'HEAD'
```
jaimerson: This could be an optional argument

jaimerson (approved):
Looks good :+1:

lib/crystalball/extensions/git/base.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Git
+  class Base # rubocop:disable Style/Documentation
+    def merge_base(from, to)
+      lib.merge_base(from, to)
+    end
```
pluff: merge-base is a commit so we should return a commit object instead of just string. See `Git::Base::Factory#commit_tree` for example. Then we can do `repo.merge_base('HEAD^", "HEAD").sha` to get commit hash.

spec/extensions/git/base_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  let(:repo) { Git.open('.') }
+
+  describe '#merge_base' do
+    subject { repo.merge_base('HEAD^', 'HEAD') }
+
+    let(:prehead) { repo.gcommit('HEAD^').sha }
+
+    it { is_expected.to eq prehead }
+  end
+end
```
pluff, line 15: I've modified this spec a bit however you should modify it more when addressing my comment above

pluff (changes requested):
When you finish addressing comment please start work on PR to `ruby-git`

lib/crystalball/extensions/git/lib.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Git
+  class Lib # rubocop:disable Style/Documentation
+    def merge_base(from, to)
```
pluff: also merge base can have more than 2 arguments

lib/crystalball/extensions/git/base.rb
EPecherkin: I think gcommit will be better

spec/extensions/git/base_spec.rb
EPecherkin: adding `.sha` to subject is not that much :)

lib/crystalball/extensions/git/lib.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    private
+
+    def prepare_merge_base_options(opts)
+      arg_opts = []
+
+      arg_opts << '--all' if opts[:all]
+      arg_opts << '--is-ancestor' if opts[:is_ancestor]
+      arg_opts << '--independent' if opts[:independent]
+      arg_opts << '--fork-point' if opts[:fork_point]
```
pluff: `opts.map { |k,v| "--#{k}" if v }.compact` is more flexible

spec/extensions/git/base_spec.rb
pluff: why do you need `sha`? Comparing by objects doesn't work?

spec/predictor_spec.rb
```diff
@@ -3,9 +3,9 @@
 require 'spec_helper'
 
 describe Crystalball::Predictor do
-  subject(:predictor) { described_class.new(instance_double('Crystalball::ExecutionMap', cases: cases), repository) }
+  subject(:predictor) { described_class.new(instance_double('Crystalball::ExecutionMap', cases: cases), repository, from: 'HEAD^', to: 'HEAD') }
```
pluff: why do we need this?

EPecherkin: You are right. we don't need this.
About https://github.com/toptal/crystalball/pull/84#discussion_r194056983
comparing objects doesn't work because we initialize these objects with different `objectish`. We need to compare `sha` here.

spec/predictor_spec.rb
```diff
@@ -5,7 +5,7 @@
 describe Crystalball::Predictor do
   subject(:predictor) { described_class.new(instance_double('Crystalball::ExecutionMap', cases: cases), repository) }
   let(:cases) { {spec_file: %w[file1.rb]} }
-  let(:repository) { instance_double('Crystalball::GitRepo', repo_path: Pathname('.')) }
+  let(:repository) { double('Crystalball::GitRepo', merge_base: double(sha: nil), repo: double(path: '/.git'), dir: double(path: './'), repo_path: Pathname('.')) }
```
pluff: this line is still not clear. we are stubbing `diff` method below, so why do you need all these changes?

EPecherkin: Oh you are right. I don't need `repo` and `dir` anymore

lib/crystalball/extensions/git/base.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Git
+  class Base # rubocop:disable Style/Documentation
```
pluff: documentation would be nice. `@return`

lib/crystalball/extensions/git/lib.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Git
+  class Lib # rubocop:disable Style/Documentation
```
pluff: same here

pluff (approved):

PR #85, opened 2018-06-06
Change limit behavior

pluff:
Current behavior of `examples_limit`is not so useful for local development because if we abort the process that means we can't do anything useful in this case. It would be more useful to write a warning about prediction size being over the limit and try to shrink prediction to specified limit size so we can provide at least some feedback.

### DIFF
--- a/lib/crystalball/rspec/examples_pruner.rb
+++ b/lib/crystalball/rspec/examples_pruner.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module Crystalball
+  module RSpec
+    # A class to prune prediction size to specified limit.
+    class ExamplesPruner
+      attr_reader :world, :to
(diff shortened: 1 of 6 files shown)


lib/crystalball/rspec/examples_pruner.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module Crystalball
+  module RSpec
+    # A class to prune prediction size to specified limit.
+    class ExamplesPruner
+      attr_reader :world, :to
+
+      def initialize(rspec_world, to:)
```
jaimerson: Please document the expected param types

jaimerson (approved):
It works great :shipit:

EPecherkin (approved):

PR #86, opened 2018-06-07
Support factory_bot

EPecherkin:


jaimerson (approved):
Looks good to me. Do you think we should support `FactoryGirl` as well? cc @EPecherkin @pluff

EPecherkin:
@jaimerson FactoryGirl was renamed to FactoryBot

jaimerson:
@EPecherkin I'm aware :smile:
My question is: should we support the old gem/constant?

EPecherkin:
@jaimerson I don't think so to be honest

EPecherkin:
@jaimerson @pluff now it supports factory_girl as well

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
```diff
@@ -0,0 +1,39 @@
...
+              alias_method :factory, :cb_original_factory # rubocop:disable Lint/DuplicateMethods
+              undef_method :cb_patched_factory
+            end
+          end
+        end
+
+        # Will replace original `FactoryBot::Syntax::Default::DSL#factory`. Pushes path of a factory to
+        # `FactoryBotStrategy.factory_definitions` and calls original `factory`
+        def cb_patched_factory(*args, &block)
+          FactoryBotStrategy.factory_definitions[args.first.to_s] = caller(1..1).first.split(':').first
```
EPecherkin: This may break if someone patched this method before us. I have no idea what we can do here

lib/crystalball/map_generator/factory_bot_strategy/factory_runner_patch.rb
```diff
@@ -0,0 +1,38 @@
...
+module Crystalball
+  class MapGenerator
+    class FactoryBotStrategy
+      # Module to add new patched `run` method to FactoryBot::FactoryRunner
+      module FactoryRunnerPatch
+        class << self
+          # Patches `FactoryBot::FactoryRunner#run`. Renames original `run` to `cb_original_run` and
+          # replaces it with custom one
+          def apply!
+            FactoryBotStrategy.factory_bot_constant::FactoryRunner.class_eval do
```
EPecherkin: This way to patch is not compatible if someone uses `Module.prepend` and tries to call an original method with `super` - it will create an endless cycle
I have an idea to use `Module.prepend` here and for all our patches to handle such case
pros: `super` works for everyone 
cons: we can't revert our patches anymore. But do we need to revert them at all?
WDYT @pluff @jaimerson ?

features/fixtures/simple_app/spec/views/index.html.erb_spec.rb
```diff
@@ -4,7 +4,7 @@
 
 describe 'index.html.erb' do
   include_context 'action view'
-  let(:assigns) { {models: [Model1.new('foo'), Model1.new('bar')]} }
+  let(:assigns) { {models: [build(:model1, field: 'foo'), build(:model1, field: 'bar')]} }
```
pluff: I think we shouldn't rely on FactoryBot while testing other strategies

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
pluff: We can specify factories directory in strategy configuration and detect the first file from specified directory. Can be `spec/factories` by default.

features/fixtures/simple_app/spec/spec_helper.rb
```diff
@@ -32,3 +35,11 @@
 require_relative '../lib/class2_eval.rb'
 require_relative '../models/model1.rb'
 require_relative '../models/model2.rb'
+
+RSpec.configure do |config|
+  config.include FactoryBot::Syntax::Methods
```
pluff, line 41: will it work without `methods` included?

lib/crystalball/map_generator/factory_bot_strategy.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require 'crystalball/map_generator/factory_bot_strategy/factory_gem_loader'
+
+Crystalball::MapGenerator::FactoryBotStrategy::FactoryGemLoader.require!
```
pluff, line 5: I don't think we need automatic gem loading for now. Other strategies rely on gems to be properly loaded by a user, so lets stick to this behavior here too.

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
```diff
@@ -0,0 +1,39 @@
...
+module Crystalball
+  class MapGenerator
+    class FactoryBotStrategy
+      # Module to add new patched `factory` method to FactoryBot::Syntax::Default::DSL
+      module DSLPatch
+        class << self
+          # Patches `FactoryBot::Syntax::Default::DSL#factory`. Renames original `factory` to `cb_original_factory` and
+          # replaces it with custom one
+          def apply!
+            FactoryBotStrategy.factory_bot_constant::Syntax::Default::DSL.class_eval do
```
pluff: https://github.com/thoughtbot/factory_bot/blob/master/lib/factory_bot/syntax/default.rb#L7-L11
I see you've patched first DSL, what about second one?

lib/crystalball/map_generator/factory_bot_strategy/factory_runner_patch.rb
```diff
@@ -0,0 +1,38 @@
...
+          def revert!
+            FactoryBotStrategy.factory_bot_constant::FactoryRunner.class_eval do
+              alias_method :run, :cb_original_run # rubocop:disable Lint/DuplicateMethods
+              undef_method :cb_patched_run
+            end
+          end
+        end
+
+        # Will replace original `FactoryBot::FactoryRunner#run`. Pushes path of a factory to
+        # `FactoryBotStrategy.factories` and calls original `run`
```
pluff: `used_factories`

lib/crystalball/map_generator/factory_bot_strategy/factory_runner_patch.rb
```diff
@@ -0,0 +1,38 @@
...
+        def cb_patched_run(*args)
+          FactoryBotStrategy.used_factories.push FactoryBotStrategy.factory_definitions[@name.to_s]
```
pluff: Given you have `FactoryBotStrategy.used_factories` interface declaration what do you expect as the return value of this method? My bet is a list of factories used. In fact, it is a magic list of file paths. Why does runner patch know anything about DSL patch logic?
```ruby
def cb_patched_run(*args)
  FactoryBotStrategy.used_factories.push @name.to_s
end

# factory_bot_strategy.rb
def call(case_map, _)
  # ...
  case_map.push(*filter(self.class.used_factories.map { |f| self.class.factory_definitions[f] }))
end
```

pluff (changes requested):

features/fixtures/simple_app/spec/spec_helper.rb
EPecherkin: It will. But I added a separate case to test it

lib/crystalball/map_generator/factory_bot_strategy.rb
EPecherkin: @pluff we don't rely on user here. Just check 
https://github.com/toptal/crystalball/blob/master/lib/crystalball/rails/map_generator/i18n_strategy/simple_patch.rb#L3
https://github.com/toptal/crystalball/blob/master/lib/crystalball/rails/map_generator/action_view_strategy/patch.rb#L3
https://github.com/toptal/crystalball/blob/master/lib/crystalball/map_generator/parser_strategy/processor.rb#L3

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
EPecherkin: I missed it. Now it is done

lib/crystalball/map_generator/factory_bot_strategy/factory_runner_patch.rb
EPecherkin: good point. fixed

lib/crystalball/map_generator/factory_bot_strategy.rb
jaimerson: Those aren't required [by default](https://github.com/toptal/crystalball/blob/master/lib/crystalball.rb#L16-L20).

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
```diff
@@ -0,0 +1,71 @@
...
+              klass.class_eval do
+                alias_method :factory, :cb_original_factory
+                undef_method :cb_patched_factory
+              end
+            end
+          end
+
+          private
+
+          def klasses_to_patch
```
jaimerson: This does not need to be `klasses`, I think `classes_to_patch` will be fine

lib/crystalball/map_generator/factory_bot_strategy/factory_runner_patch.rb
jaimerson: I think reverting the patches is necessary to not have unintended consequences on the next strategies. Do we have this issue with `prepend` right now? If not I think we can keep it this way.

lib/crystalball/map_generator/factory_bot_strategy.rb
EPecherkin: Just like this one, isn't it?

lib/crystalball/map_generator/factory_bot_strategy/factory_runner_patch.rb
EPecherkin: We have it on another project. Exactly the case I described

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
```diff
@@ -0,0 +1,53 @@
...
+
+          FactoryBotStrategy.factory_definitions[name] ||= []
+          FactoryBotStrategy.factory_definitions[name] << factory_path
+
+          super(*args, &block)
+        end
+
+        private
+
+        def cb_fetch_factory_path
```
pluff: We should try to avoid polluting objects with our methods if possible. This method can be extracted to separate service method.

features/fixtures/simple_app/spec/models/model1_spec.rb
```diff
@@ -10,8 +10,20 @@
   end
 
   describe '#field' do
-    subject { described_class.new('value').field }
+    subject { described_class.new(field: 'value').field }
 
     it { is_expected.to eq 'value' }
   end
+
+  context 'using factory_bot with syntax included' do
```
pluff: same context descriptions

lib/crystalball/map_generator/factory_bot_strategy.rb
```diff
@@ -0,0 +1,58 @@
...
+        DSLPatch.apply!
+        FactoryRunnerPatch.apply!
+      end
+
+      # Adds factories related to the spec to the case map
+      # @param [Crystalball::CaseMap] case_map - object holding example metadata and affected files
+      def call(case_map, _)
+        self.class.reset_used_factories
+        yield case_map
+        case_map.push(*filter(self.class.used_factories.flat_map { |f| self.class.factory_definitions[f] }.compact))
```
pluff: what's the case when we have no factory definition but the factory was used?

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
```diff
@@ -0,0 +1,53 @@
...
+
+          def classes_to_patch
+            [
+              FactoryBotStrategy.factory_bot_constant::Syntax::Default::DSL,
+              FactoryBotStrategy.factory_bot_constant::Syntax::Default::ModifyDSL
+            ]
+          end
+        end
+
+        # Will replace original `FactoryBot::Syntax::Default::DSL#factory` and `FactoryBot::Syntax::Default::ModifyDSL#factory`.
```
pluff: `Overrides`

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
```diff
@@ -0,0 +1,53 @@
...
+        # Pushes path of a factory to `FactoryBotStrategy.factory_definitions` and calls original `factory`
+        def factory(*args, &block)
+          factory_path = cb_fetch_factory_path
+          name = args.first.to_s
```
pluff: you can call for `super` first and then rely on object returned instead of rely on method interface which can be a subject to change by factory bot authors.

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
```diff
@@ -0,0 +1,53 @@
...
+        end
+
+        private
+
+        def cb_fetch_factory_path
+          call = caller.find do |c|
+            factories_definition_paths = FactoryBotStrategy.factory_bot_constant.definition_file_paths
+
+            factories_definition_paths.any? do |path|
+              full_path = Pathname(path).expand_path.to_s
```
pluff: you can move it out of iterations and memoize.

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
```diff
@@ -0,0 +1,53 @@
...
+          FactoryBotStrategy.factory_definitions[name] ||= []
+          FactoryBotStrategy.factory_definitions[name] << factory_path
+
+          super(*args, &block)
+        end
+
+        private
+
+        def cb_fetch_factory_path
+          call = caller.find do |c|
```
pluff: `factory_definition_file`

lib/crystalball/map_generator/factory_bot_strategy/factory_runner_patch.rb
```diff
@@ -0,0 +1,24 @@
...
+          def apply!
+            FactoryBotStrategy.factory_bot_constant::FactoryRunner.prepend FactoryRunnerPatch
+          end
+        end
+
+        # Will replace original `FactoryBot::FactoryRunner#run`. Pushes factory name to
+        # `FactoryBotStrategy.used_factories` and calls original `run`
+        def run(*args, &block)
+          FactoryBotStrategy.used_factories << @name.to_s
+          super(*args, &block)
```
pluff: `super`

lib/crystalball/map_generator/factory_bot_strategy/factory_runner_patch.rb
```diff
@@ -0,0 +1,24 @@
...
+        class << self
+          # Patches `FactoryBot::FactoryRunner#run`.
+          def apply!
+            FactoryBotStrategy.factory_bot_constant::FactoryRunner.prepend FactoryRunnerPatch
+          end
+        end
+
+        # Will replace original `FactoryBot::FactoryRunner#run`. Pushes factory name to
+        # `FactoryBotStrategy.used_factories` and calls original `run`
+        def run(*args, &block)
```
pluff: `*`

spec/map_generator/factory_bot_strategy/dsl_patch_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'crystalball/factory_bot'
+
+describe Crystalball::MapGenerator::FactoryBotStrategy::DSLPatch do
+  subject(:instance) do
+    Class.new do
+      def factory(*args, &block)
+        "original factory called with #{args}" + (' and block' if block)
```
pluff: you can simply retrun args and block from this method

lib/crystalball/map_generator/factory_bot_strategy.rb
```diff
@@ -0,0 +1,58 @@
...
+      def after_register
+        DSLPatch.apply!
+        FactoryRunnerPatch.apply!
+      end
+
+      # Adds factories related to the spec to the case map
+      # @param [Crystalball::CaseMap] case_map - object holding example metadata and affected files
+      def call(case_map, _)
+        self.class.reset_used_factories
+        yield case_map
```
pluff: Please update to latest variable names. Also please pass all arguments to block

jaimerson (approved):

lib/crystalball/map_generator/factory_bot_strategy/dsl_patch.rb
EPecherkin: It returns not a factory
https://github.com/thoughtbot/factory_bot/blob/master/lib/factory_bot/syntax/default.rb#L22

features/fixtures/simple_app/spec/models/model1_spec.rb
```diff
@@ -15,14 +15,14 @@
...
   end
 
-  context 'using factory_bot with syntax included' do
+  context 'using factory_bot without syntax included' do
     subject { FactoryBot.create(:model1) }
 
     it { is_expected.to have_attributes(name: 'John Doe', field: 'value') }
   end
 
   context 'using factory_bot with syntax included' do
```
jaimerson: This context should be `when using a factory alias`

jaimerson (approved):

lib/crystalball/map_generator/factory_bot_strategy.rb
EPecherkin: When it was used by an alias. Now I solved that case

pluff (approved):

PR #87, opened 2018-06-13
Fix a bug with prediction being empty

pluff:
if there is no predicted example group that fits the limit

+ Fix a 🐛 with filtering logic and pure file predictions


lib/crystalball/rspec/prediction_pruning/examples_pruner.rb
```diff
@@ -0,0 +1,64 @@
...
+              prune_to_limit(g)
+            end
+
+            add_examples(group)
+          else
+            add(group, group_size)
+          end
+        end
+
+        def add(group_or_example, group_size)
```
jaimerson: This method name is not descriptive. Should be something like `add_group_or_example_to_result`. Also `group_size` can default to `1`, making the addition of a single example or group read better.

lib/crystalball/rspec/runner.rb
```diff
@@ -70,44 +72,21 @@ def config_file
...
+      def setup(err, out)
+        super
 
-        out.puts "Prediction examples size #{@world.example_count(pruner.world_groups)} is over the limit (#{limit})"
-        out.puts "Prediction is pruned to fit the limit!"
+        return unless reconfiguration_needed?
 
-        pruner.pruned_groups
+        reconfigure_to_limit
+        super
```
jaimerson: Why are you calling `super` twice?

spec/rspec/prediction_pruning/examples_pruner_spec.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Crystalball::RSpec::PredictionPruning::ExamplesPruner do
+  subject(:pruner) { described_class.new(world, to: limit) }
+
+  let(:world) { instance_double('RSpec::Core::World') }
+  let(:limit) { 100 }
```
jaimerson: I don't think this is needed. It's set on all the contexts below.

lib/crystalball/rspec/prediction_pruning/examples_pruner.rb
```diff
@@ -0,0 +1,64 @@
...
+    module PredictionPruning
+      # A class to prune given world example groups to fit the limit.
+      class ExamplesPruner
+        attr_reader :world, :to
+
+        # @param [RSpec::Core::World] rspec_world RSpec world instance
+        # @param [Integer] to upper bound limit for prediction.
+        def initialize(rspec_world, to:)
+          @world = rspec_world
+          @to = to
```
jaimerson: The keyword argument `to:` is cool, but you can give a more meaningful name to the ivar, like `limit`, `examples_limit` or `upper_bound_limit`.

jaimerson (changes requested):
Also remember to change [this](https://github.com/toptal/crystalball/blob/master/features/rspec_runner_spec.rb#L24) back to `1 example, 0 failures`

spec/rspec/runner_spec.rb
```diff
@@ -114,37 +115,40 @@
...
       end
+      after { ENV.delete('CRYSTALBALL_EXAMPLES_LIMIT') }
 
-      after { ENV.delete('CRYSTALBALL_CONFIG') }
-
-      it 'returns whatever ExamplesPruner returns' do
-        pruner = double(world_groups: %w[a b], pruned_groups: ['pruned_groups'])
-        allow(Crystalball::RSpec::ExamplesPruner).to receive(:new).with(world, to: 1).and_return pruner
+      it 'reconfigures RSpec env with new set from ExamplesPruner' do
+        pruner = double(pruned_set: ['pruned_set'])
```
EPecherkin: use `let` for this
and you can place both `allow`s into `before` block

lib/crystalball/rspec/runner.rb
pluff: Because RSpec needs to be reconfigured again after changing filters. I reworked a bit so there is no `super` calls anymore

spec/rspec/runner_spec.rb
pluff: `let` doesn't make any sense since it is not something that should be reused.

spec/rspec/filtering_spec.rb
```diff
@@ -16,6 +16,16 @@
       ).as_null_object
     end
 
+    context 'when filter manager has no inclusions defined' do
+      let(:files) { %w[./spec/foo_spec.rb] }
+      let(:inclusions) { {} }
+
+      it 'should not change any filtering rules' do
+        remove_unnecessary_filters
+        expect(inclusions[:ids]).to be_nil
```
jaimerson: This is testing that the hash created in line 21 is the hash created in line 21. Probably better to do something like
```ruby
expect { remove_unecessary_filters }.not_to raise_error
```

pluff: it tests that the hash wasn't changed.

jaimerson (approved):
:shipit:

spec/rspec/filtering_spec.rb
```diff
@@ -16,6 +16,16 @@
       ).as_null_object
     end
 
+    context 'when filter manager has no inclusions defined' do
+      let(:files) { %w[./spec/foo_spec.rb] }
+      let(:inclusions) { {} }
+
+      it 'does not change any filtering rules' do
+        remove_unnecessary_filters
+        expect(inclusions[:ids]).to be_nil
```
EPecherkin: If you want to test that `remove_unnecessary_filters` do not modify `inclusions[:ids]` you can use more descriptive syntax
`expect { remove_unnecessary_filters }.not_to change { inclusions[:ids] }`

EPecherkin (approved):

PR #88, opened 2018-06-14
Unify logging

jaimerson:


lib/crystalball/logger.rb
```diff
@@ -0,0 +1,75 @@
...
+      error: "\e[31m",
+      critical: "\e[101m"
+    }.freeze
+
+    include Singleton
+
+    class << self
+      extend Forwardable
+
+      delegate %i[
```
EPecherkin: I think logger should be used as `Crystalball.logger` only, without class level methods
I don't really like an idea to use `Crystalball::Logger` to log anything
what do you think @pluff ?

lib/crystalball/rails/predictor/modified_schema.rb
```diff
@@ -48,6 +48,10 @@ def tables_map
 
         private
 
+        def logger
+          Crystalball::Logger
```
EPecherkin: no needs for such methods if you use `Crystallball.logger`

lib/crystalball/rspec/runner.rb
```diff
@@ -69,44 +71,55 @@ def config_file
           file.exist? ? file : nil
         end
 
-        def build_prediction(out)
-          check_map(out)
+        def setup_logger
+          Crystalball::Logger.configure do |logger|
+            logger.log_level = config['log_level']
+            logger.color = config['color'].to_s == 'true' # ENV vars can only be string
```
EPecherkin: just define a proper getter for this variable in `lib/crystalball/rspec/runner/configuration.rb`

spec/rspec/runner_spec.rb
```diff
@@ -106,11 +118,9 @@
     context 'with expired map' do
       before { allow(prediction_builder).to receive(:expired_map?).and_return true }
 
-      let(:out_stream) { double(puts: true) }
-
       it 'prints out warning' do
-        expect(out_stream).to receive(:puts).with('Maps are outdated!')
-        described_class.run([], STDERR, out_stream)
+        expect(Crystalball::Logger).to receive(:warn).with('Maps are outdated!')
```
EPecherkin: your stubs are inconsistent [here](https://github.com/toptal/crystalball/pull/88/files#diff-21ff99f1c30317f4701f73bebb218093R100) and [here](https://github.com/toptal/crystalball/pull/88/files#diff-21ff99f1c30317f4701f73bebb218093R27) you stub it as `Crystalball::Logger.instance`

EPecherkin (changes requested):

lib/crystalball/logger.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'singleton'
+
+module Crystalball
+  # This class is responsible for logging information to the configured output stream.
+  class Logger
```
pluff: why don't you reuse https://ruby-doc.org/stdlib-2.1.0/libdoc/logger/rdoc/Logger.html ?

lib/crystalball/rspec/runner.rb
```diff
@@ -11,17 +11,18 @@ module RSpec
...
           end
 
-          out.puts "Crystalball starts to glow..."
-          prediction = build_prediction(out)
+          logger.info "Crystalball starts to glow..."
+          prediction = build_prediction
 
-          out.puts "Prediction: #{prediction.first(5).join(' ')}#{'...' if prediction.size > 5}"
-          out.puts "Starting RSpec."
+          logger.info "Prediction: #{prediction.first(5).join(' ')}#{'...' if prediction.size > 5}"
```
pluff: I think this can be `debug`

lib/crystalball/rspec/runner.rb
```diff
@@ -69,44 +71,55 @@ def config_file
...
-        def prune_prediction_to_limit(prediction, out)
+        def prune_prediction_to_limit(prediction)
           limit = config['examples_limit'].to_i
 
           return prediction if !limit.positive? || prediction.size <= limit
 
-          out.puts "Prediction size #{prediction.size} is over the limit (#{limit})"
-          out.puts "Prediction is pruned to fit the limit!"
+          logger.info "Prediction size #{prediction.size} is over the limit (#{limit})"
+          logger.info "Prediction is pruned to fit the limit!"
```
pluff: these are warnings

lib/crystalball/rspec/runner.rb
```diff
@@ -69,44 +71,55 @@ def config_file
...
         limit = self.class.config['examples_limit'].to_i
 
         pruner = ExamplesPruner.new(@world, to: limit)
 
         return pruner.world_groups if !limit.positive? || @world.example_count(pruner.world_groups) <= limit
 
-        out.puts "Prediction examples size #{@world.example_count(pruner.world_groups)} is over the limit (#{limit})"
-        out.puts "Prediction is pruned to fit the limit!"
+        Crystalball::Logger.info "Prediction examples size #{@world.example_count(pruner.world_groups)} is over the limit (#{limit})"
+        Crystalball::Logger.info "Prediction is pruned to fit the limit!"
```
pluff: these are also warnings

spec/rspec/runner/configuration_spec.rb
```diff
@@ -61,7 +63,9 @@
           'diff_from' => 'HEAD~3',
           'diff_to' => 'HEAD',
           'map_expiration_period' => 1,
-          'custom' => 42
+          'custom' => 42,
```
pluff: let's keep `custom` as last. because it is custom :)

lib/crystalball/rspec/runner.rb
```diff
@@ -11,17 +11,18 @@ module RSpec
...
+          setup_logger
           return config['runner_class'].run(args, err, out) unless config['runner_class'] == self
 
           ::RSpec.configure do |c|
             c.silence_filter_announcements = true
           end
 
-          out.puts "Crystalball starts to glow..."
-          prediction = build_prediction(out)
+          logger.info "Crystalball starts to glow..."
```
pluff: I guess we'll need 2 logger instances one for STDOUT and one for `log/crystalball.log`

lib/crystalball/rspec/runner.rb
jaimerson: Hm, nice, hadn't thought about that

spec/rspec/runner_spec.rb
jaimerson: That's because `Logger.configure` does `instance_eval` on the instance, so it's the instance that receives the `log_level=` for example. I think I can change it all to expect on the instance, for the sake of consistency.

pluff (changes requested):

lib/crystalball/logger.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'logger'
+
+module Crystalball
+  class << self
+    # This logs information to the standard output based on the configured log level,
+    # and also logs unfiltered information to the configured log file.
+    def log(severity, message)
```
pluff: https://ruby-doc.org/stdlib-2.1.0/libdoc/logger/rdoc/Logger.html#method-i-add

lib/crystalball/logger.rb
```diff
@@ -0,0 +1,51 @@
...
+    private
+
+    LOG_LEVELS = {
+      debug: ::Logger::DEBUG,
+      info: ::Logger::INFO,
+      warn: ::Logger::WARN,
+      error: ::Logger::ERROR,
+      fatal: ::Logger::FATAL,
+      unknown: ::Logger::UNKNOWN
+    }.freeze
```
pluff:
```ruby
def severity(severety_sym)
  ::Logger.const_get(severy_sym.to_s.upcase)
end
```

lib/crystalball/logger.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def output_stream
+      @output_stream ||= ::Logger.new(STDOUT).tap do |logger|
+        logger.level = LOG_LEVELS.fetch(configured_level)
+      end
+    end
+
+    def log_file_output_stream
+      @log_file_stream ||= ::Logger.new(config['log_file']).tap do |logger|
```
pluff: What if I don't want to write log file?

spec/logger_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    let(:configured_level) { 'unknown' }
+
+    include_examples 'log file behavior'
+
+    it 'logs only unknown' do
+      result = output_stream.string.gsub("\n", '')
+      expect(result).not_to match(/.*DEBUG.*INFO.*WARN.*ERROR.*FATAL.*/)
+      expect(result).to match(/.*UNKNOWN/)
+    end
+  end
```
pluff: You don't have to test Logger constants here. One test with some severity as STDOUT logger level is enough.

spec/rspec/runner_spec.rb
```diff
@@ -11,6 +11,7 @@
     described_class.reset!
     allow(Crystalball::MapStorage::YAMLStorage).to receive(:load).and_return(map)
     allow(RSpec::Core::ExampleGroup).to receive(:run)
+    allow(Crystalball).to receive(:log)
```
pluff: why do you need it here?

spec/rspec/runner/configuration_spec.rb
```diff
@@ -19,7 +19,9 @@
           'diff_from' => 'HEAD',
           'diff_to' => nil,
           'runner_class' => Crystalball::RSpec::Runner,
-          'runner_class_name' => 'Crystalball::RSpec::Runner'
+          'runner_class_name' => 'Crystalball::RSpec::Runner',
+          'log_level' => :info,
+          'log_file' => 'tmp/crystalball.log'
```
pluff: this can be `log/crystalball.log` by default

pluff (changes requested):

lib/crystalball/logger.rb
jaimerson: Set `/dev/null` in config

pluff (approved):

lib/crystalball/logger.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require 'logger'
+
+module Crystalball
+  class << self
```
EPecherkin: Make it a separate module and extend `Crystalball` module with it

lib/crystalball/logger.rb
```diff
@@ -0,0 +1,46 @@
...
+    # This logs information to the standard output based on the configured log level,
+    # and also logs unfiltered information to the configured log file.
+    def log(severity_sym, *args, &block)
```
EPecherkin: where are tests for this method?

EPecherkin (changes requested):

lib/crystalball/logger.rb
jaimerson: https://github.com/toptal/crystalball/blob/d95d623c62af37a0d26c994e04c91ca881de263d/spec/logger_spec.rb

lib/crystalball/logger.rb
jaimerson: Would you care to clarify the point? It would have the same outcome with different syntax, and would not be extended elsewhere.

EPecherkin (approved):

PR #92, opened 2018-06-21
Use a different map generator configuration for each feature

jaimerson:
### Usage
```ruby
# features/my_shiny_feature_spec.rb
require 'feature_helper'

RSpec.describe 'My feature with map generator' do
  include_context 'simple git repository'
  let(:map_generator_config) do
    # must return a string containing the code that will be written in spec_helper.rb
    <<~CONFIG
      end
      Crystalball::Rails::TablesMapGenerator.start! # if it also needs a table map
    CONFIG
  end
  # ...
end
```

### DIFF
--- a/features/evaluation_spec.rb
+++ b/features/evaluation_spec.rb
@@ -9,6 +9,14 @@ describe 'Prediction evaluation' do
   end
   include_context 'simple git repository'
 
+  map_generator_config do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
+        c.register Crystalball::MapGenerator::CoverageStrategy.new
+      end
+    CONFIG
+  end
+
   let(:predictor) do
     Crystalball::Predictor.new(map, Crystalball::GitRepo.open(git.dir.path)) do |predictor|
       predictor.use Crystalball::Predictor::AssociatedSpecs.new from: %r{models/(?<file>.*).rb},
--- a/features/modified_execution_paths/change_in_branch_spec.rb
+++ b/features/modified_execution_paths/change_in_branch_spec.rb
@@ -5,11 +5,27 @@ require_relative '../feature_helper'
 describe 'Changing source file in a branch' do
   include_context 'simple git repository'
   include_context 'class1 examples'
-  include_context 'model1 examples'
   include_context 'base forecast'
+
+  map_generator_config do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
+        c.register Crystalball::MapGenerator::DescribedClassStrategy.new
+      end
+    CONFIG
+  end
 
   let(:strategies) do
     [Crystalball::Predictor::ModifiedExecutionPaths.new]
+  end
+
+  let(:model1_examples) do
+    [
+      './spec/models/model1_spec.rb[1:1:1]',
+      './spec/models/model1_spec.rb[1:2:1]',
+      './spec/models/model1_spec.rb[1:3:1]',
+      './spec/models/model1_spec.rb[1:4:1]'
+    ]
   end
 
   before do
--- a/features/modified_schema/change_spec.rb
+++ b/features/modified_schema/change_spec.rb
@@ -2,10 +2,29 @@
 
 require_relative '../feature_helper'
 
-describe 'Changing schema file' do
+RSpec.describe 'Changing schema file' do
   include_context 'simple git repository'
-  include_context 'model1 examples'
   include_context 'base forecast'
+  let(:schema_path) { root.join('db', 'schema.rb') }
+
+  map_generator_config do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
+        c.register Crystalball::MapGenerator::DescribedClassStrategy.new
+      end
+
+      Crystalball::Rails::TablesMapGenerator.start!
+    CONFIG
+  end
+
+  let(:model1_examples) do
+    [
+      './spec/models/model1_spec.rb[1:1:1]',
+      './spec/models/model1_spec.rb[1:2:1]',
+      './spec/models/model1_spec.rb[1:3:1]',
+      './spec/models/model1_spec.rb[1:4:1]'
+    ]
+  end
 
   let(:strategies) do
     [Crystalball::Rails::Predictor::ModifiedSchema.new(tables_map_path: root.join('tables_map.yml'))]
--- a/features/rspec_runner_spec.rb
+++ b/features/rspec_runner_spec.rb
@@ -10,6 +10,14 @@ describe 'RSpec runner' do
   let(:important_class_path) { root.join('lib/important_class.rb') }
   let(:other_important_class_path) { root.join('lib/other_important_class.rb') }
 
+  map_generator_config do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
+        c.register Crystalball::MapGenerator::CoverageStrategy.new
+      end
+    CONFIG
+  end
+
   before do
     ENV['CRYSTALBALL_LOG_LEVEL'] = 'debug'
     ENV['CRYSTALBALL_LOG_FILE'] = '/dev/null'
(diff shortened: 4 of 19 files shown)


features/evaluation_spec.rb
```diff
@@ -9,6 +9,21 @@
   end
   include_context 'simple git repository'
 
+  let(:map_generator_config) do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
```
EPecherkin: I think we should use CoverageStrategy only for this feature

features/evaluation_spec.rb
```diff
@@ -27,6 +42,7 @@
       class Model1
       end
     RUBY
+    git.checkout(spec_helper)
```
EPecherkin: Let's commit this changes before we run anything. So, we will be on a safe side forever

features/modified_execution_paths/change_in_branch_spec.rb
```diff
@@ -8,6 +8,16 @@
   include_context 'model1 examples'
   include_context 'base forecast'
 
+  let(:map_generator_config) do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
+        c.register Crystalball::MapGenerator::CoverageStrategy.new
+        c.register Crystalball::MapGenerator::DescribedClassStrategy.new
+        c.register Crystalball::MapGenerator::AllocatedObjectsStrategy.build(only: ['Object'])
```
EPecherkin: I think it is not necessary to have all of these here. We can stay with `CoverageStrategy` only

features/modified_execution_paths/change_spec.rb
```diff
@@ -8,6 +8,16 @@
   include_context 'model1 examples'
   include_context 'base forecast'
 
+  let(:map_generator_config) do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
+        c.register Crystalball::MapGenerator::CoverageStrategy.new
+        c.register Crystalball::MapGenerator::AllocatedObjectsStrategy.build(only: ['Object'])
```
EPecherkin: and here as well. we can go with `CoverageStrategy` only

features/modified_schema/change_spec.rb
```diff
@@ -2,10 +2,23 @@
...
+RSpec.describe 'Changing schema file' do
   include_context 'simple git repository'
   include_context 'model1 examples'
   include_context 'base forecast'
+  let(:schema_path) { root.join('db', 'schema.rb') }
+
+  let(:map_generator_config) do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
+        c.register Crystalball::MapGenerator::CoverageStrategy.new
```
EPecherkin: `TablesMapGenerator` + `CoverageStrategy` should be enough I think

features/rspec_runner_spec.rb
```diff
@@ -10,6 +10,21 @@
   let(:important_class_path) { root.join('lib/important_class.rb') }
   let(:other_important_class_path) { root.join('lib/other_important_class.rb') }
 
+  let(:map_generator_config) do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
+        c.register Crystalball::MapGenerator::CoverageStrategy.new
+        c.register Crystalball::MapGenerator::AllocatedObjectsStrategy.build(only: ['Object'])
```
EPecherkin: We can drop everything except `CoverageStrategy` as well I think

features/support/shared_contexts/simple_git_repository.rb
```diff
@@ -8,24 +8,21 @@
...
   let(:spec_path) { root.join('spec') }
   let(:class1_spec_path) { spec_path.join('class1_spec.rb') }
-  let(:factories_path) { spec_path.join('factories') }
-  let(:model1_factory_path) { factories_path.join('model1s.rb') }
-  let(:model1_factory_modification_path) { factories_path.join('model1s_modification.rb') }
   let(:action_view_shared_context) { spec_path.join('support', 'shared_contexts', 'action_view.rb') }
   let(:git) { Git.init(root.to_s) }
+  let(:spec_helper) { File.join(root, 'spec/spec_helper.rb') }
+
+  # By default, generate map just with metadata
```
EPecherkin: `generate map with metadata only` sound a bit better I think. but it is a stylish note, you can ignore it if you like original variant more :)

features/support/shared_contexts/simple_git_repository.rb
```diff
@@ -34,13 +31,19 @@
...
+    raise "Can't generate map" unless generate_map
   end
 
   after do
     root.rmtree
   end
 
+  def generate_map
+    config = map_generator_config.to_s.chomp
+    replace(spec_helper, /# MAP_GENERATOR_CONFIG/, config)
```
EPecherkin: here you change `spec_helper` and generate a map
let's split this logic
it is better to change `spec_helper` and commit it separately. so, main logic of a feature will be executed on a clean repo without unexpected side-effects

EPecherkin (dismissed):
1. commit changed `spec_helper`, so we will have a clean repo at the beginning of each feature
2. use `CoverageStrategy` only unless you want to test another strategy. it will make features more granular

features/support/shared_contexts/simple_git_repository.rb
jaimerson: You won't believe it, but I wrote exactly that and changed to this :smile:
I'm fine with either, so I'll change it.

features/evaluation_spec.rb
jaimerson: Done

features/modified_execution_paths/change_spec.rb
```diff
@@ -8,6 +8,15 @@
   include_context 'model1 examples'
   include_context 'base forecast'
 
+  let(:map_generator_config) do
+    <<~CONFIG
+      Crystalball::MapGenerator.start! do |c|
+        c.register Crystalball::MapGenerator::AllocatedObjectsStrategy.build(only: ['Object'])
+        c.register Crystalball::MapGenerator::DescribedClassStrategy.new
```
jaimerson: Here we need `AllocatedObjects` to detect `Class2.new` in `file_spec.rb` and `DescribedClass` for the `class1` examples.

pluff: Lets split it into 2 different features one for each strategy

features/evaluation_spec.rb
```diff
@@ -9,6 +9,14 @@
   end
   include_context 'simple git repository'
 
+  let(:map_generator_config) do
```
pluff:
```
map_generator_config do
  "STRING"
end
```
will be a bit more declarative

features/modified_execution_paths/change_spec.rb
```diff
@@ -84,49 +96,87 @@
...
-                                      './spec/models/model1_spec.rb[1:4:1]'
-                                    ])
+  context 'when dealing with factories' do
+    let(:map_generator_config) do
+      <<~CONFIG
+        Crystalball::MapGenerator.start! do |c|
+          c.register Crystalball::MapGenerator::FactoryBotStrategy.new
+        end
+      CONFIG
+    end
```
pluff: lets split features for `ModifiedExecutuonPaths` prediction strategy by execution strategy too.
 E.g. `features/modified_execution_paths/factory_boty_strategy_spec.rb` or `features/modified_execution_paths/factory_boty_strategy/*_spec.rb`

pluff (changes requested):

features/evaluation_spec.rb
jaimerson: I don't see much value in it, especially with the added complexity of handling overriding in nested contexts.

pluff: there is no complexity. its literally more explicit alias.
```
def map_generator_config(&block)
  let(:map_generator_config, &block)
end
```

pluff (approved):

EPecherkin (approved):

PR #95, opened 2018-08-07
Adjusts default location for mapping file

mwhagedorn:
The "out of the box" configuration wants to see the mapping file at the root of the project.  This seems like a more sensible default.


pluff:
@mwhagedorn we try to guess the best place for mappings here. `tmp` folder is in `.gitignore` usually, that makes it more convenient to use as a default. Your changes will add one more step before a user can use the tool.

PR #104, opened 2018-12-12
Handle execution maps without example groups

noralin:
We're parallelizing our tests on CI and generating multiple execution maps. 

Sometimes we don't actually run any tests on a specific worker, but an execution map is still generated without any example groups.

We ran into `no implicit conversion of nil into Hash`, so this PR compacts the example groups before merging them.

cc @MarkyMarkMcDonald


pluff:
@noralin sorry for delay. Thanks for your contribution it will be available in next release.

PR #108, opened 2019-03-18
Fix SQLite Gem to 1.3.13

kbruccoleri:
The feature specs run rspec using a system call to determine whether or not they execution maps were being generated properly. By running the call and not piping it to dev/null, it seems that rspec was failing to run on the sample app due to a gem version misalignment between ActiveRecord and sqlite due to the recent release of sqlite 1.4.0. To solve the problem, I fixed the version to the most recent of sqlite that allowed the feature tests to pass.


kbruccoleri:
@pluff

pluff:
👍 merged

PR #126, opened 2020-02-10
Fix Simplecov integration

henrich-m:
Why? Since the 0.18 version, `Simplecov.start` cannot be called twice.
So we only need to start Simplecov when it isn't running.


joshRpowell:
@pirj pinging as your the last commit to master. 

this worked for us. Simplecov is on 0.19. Can this get merged and released?

pirj:
Unfortunately, I don't have merge rights.
Summoning @jaimerson and @EPecherkin ⚡

etagwerker (approved):
This looks good to me. 💯

Issue #117, opened 2019-07-11
Get prediction as a list in STDOUT 


pirj:
Crystalball needs to be able to list predicted examples without actually running them. It's useful for further load balancing with Knapsack\whatever.

Proposed interface:
`crystalball --dry-prediction` which outputs a list of predicted examples one per line.

Related to #112

henrich-m:
It would also make Cucumber Support a lot easier

Neal910:
Any update on this? or alternative ways? Thanks.

pirj:
A pull request is welcome!

Neal910:
Hi @pirj , I would like to work on this issue although I am super new to the Ruby world. I would like to make a quick WIP PR, but doesn't seem to have permission to push to a new branch. Can you help me out? Thanks.

Neal910:
With my prototype PR above, I have been able to output the prediction lists. However per each path the format is something like this `./spec/models/**/create_line_items_spec.rb[1:1:1:1:2]`.
Is there a way to make the output using line number instead of this [1:1:1:2] structure? Because line number is accepted by most of the rspec framework.

pirj:
Couldn't quickly find it in the [docs](https://relishapp.com/rspec/rspec-core/v/3-9/docs/command-line/line-number-appended-to-file-path), but [RSpec understands `[1:1:1:1:2]` as well](https://github.com/rspec/rspec-core/blob/c095d6dac68b76ac85957dc417751f014cccf316/spec/integration/filtering_spec.rb#L238).

pluff:
Spec address based on line number doesn't work well with shared examples or cycles, that's why 🔮 uses example id.

PR #129, opened 2020-08-27
Add dry-run option

Neal910:
This PR is meant to tackle the issue of https://github.com/toptal/crystalball/issues/117

Background:
I would like to output the prediction of a list of Rspec tests, so that the [Knapsack Gem](https://docs.knapsackpro.com/ruby/knapsack) can accept the string and run the tests in parallel, as described [here](https://knapsackpro.com/faq/question/how-to-run-a-specific-list-of-test-files-or-only-some-tests-from-test-file).

Personally, I came from Java world and am really new to the Ruby, actually this is my first Ruby project. Therefore, I am creating this prototype PR so that I can get more feedback in the beginning. Again, this PR is not meant to be merged but to discuss my proposal. All comments and suggestions are welcoming.


pirj:
@EPecherkin @jaimerson @pluff I don't have the capacity for CrystalBall right now, appreciate if you could take this over.

Neal910:
any suggestions are appreciated. Thanks guys.

lib/crystalball/rspec/runner.rb
```diff
@@ -17,13 +17,23 @@ def run(args, err = $stderr, out = $stdout)
 
           Crystalball.log :info, "Crystalball starts to glow..."
           prediction = build_prediction
-
+          dry_prediction(prediction)
```
jaimerson: We should not proceed after printing the prediction if the flag is passed, so it would be something like
```ruby
if dry_run?
  puts prediction.to_a
  exit
end
```

lib/crystalball/rspec/runner.rb
```diff
@@ -17,13 +17,23 @@ def run(args, err = $stderr, out = $stdout)
...
+          
           Crystalball.log :debug, "Prediction: #{prediction.first(5).join(' ')}#{'...' if prediction.size > 5}"
           Crystalball.log :info, "Starting RSpec."
 
           super(args + prediction, err, out)
         end
 
+        def dry_prediction (prediction)
```
jaimerson: This method could be just the detection part, and could be called `dry_run?` (as well as have the tag named `--dry-run`) for consistency with similar command line tools.

lib/crystalball/rspec/runner.rb
```diff
@@ -17,13 +17,23 @@ def run(args, err = $stderr, out = $stdout)
...
+          args = Hash[ ARGV.flat_map{|s| s.scan(/--?([^=\s]+)(?:=(\S+))?/) } ]
+          if args.key?('dry-prediction')
+            array = prediction.to_a 
+            File.write('./crystalball_spec_list.txt', array.join(","))
```
jaimerson: It should write to STDOUT and the user would decide whether or not to redirect output to a file. Not sure if it would work with how the output is handled currently, in other places.

Neal910:
With this prototype PR, I have been able to output the prediction lists. However per each test case path the format is something like this `./spec/models/**/create_line_items_spec.rb[1:1:1:1:2]`
Is there a way to make the output using line number instead of this `[1:1:1:2]` structure? Because line number is accepted by most of the rspec framework.

lib/crystalball/rspec/runner.rb
Neal910: I made slightly change on it, hopefully that what you meant.

Neal910:
so far it's working with `--dry-run` , it's printing out the rspec list.  Then I use `sed` to manipulate the output to be acceptable to Knapsack tools.

jacob-s-son:
Hey, folks! I am planning to work on a POC next week to implement selective running of RSPEC and Cucumber tests that are load balanced by Knapsack Pro. Any guidance on the status of this PR? Any reason it can't be merged into mainstream? Maybe something I can help with?

Issue #118, opened 2019-07-11
Create flexible map generation filtering logic


pirj:
Currently we have filtering logic literally hardcoded here: [`lib/crystalball/map_generator/helpers/path_filter.rb`](https://github.com/toptal/crystalball/blob/master/lib/crystalball/map_generator/helpers/path_filter.rb).

We need it to be a bit more flexible so we can specify a list of directories\files to filter out of profiling data. 

Common use case: `bundle install --standalone` installs everything to `vendor` => falls into the data.

Expected configuration:

```ruby
Crystalball::MapGenerator.start!(exclude_sources: [/\.gems/, /vendor/], exclude_specs: [/features/]) do
  # ...
end
```

it should exclude all sources containing `.gems` or `vendor` in paths and it should exclude the data for all spec files including `features` in paths

rodrigovilina:
Is no one working on this? Perhaps I can give it a try! Is the proposed example the way I should approach it :ok_hand: ?

edit:
Looking at the code I think it would be cleaner to implement this at the strategy level, so only `PathFilter` and the strategies are modified. If I attempt to do it from the generator itself, I have to pass the arguments thru the `MapGenerator::Configuration` and then write the attribute into the strategies inside the `MapGenerator::Configuration#register` method.

edit:
I'll put up a PR implementing just how the example looks with comments on how I think it could be simpler if it is done right at the strategies level. Also this provides further granularity.

vaporyhumo:
Looking at the code I think it would be cleaner to implement this at the strategy level, so only `PathFilter` and the strategies are modified. If I attempt to do it from the generator itself, I have to pass the arguments thru the `MapGenerator::Configuration` and then write the attribute into the strategies inside the `MapGenerator::Configuration#register` method.

vaporyhumo:
I'll put up a PR implementing just how the example looks with comments on how I think it could be simpler if it is done right at the strategies level. Also this provides further granularity.

PR #131, opened 2020-12-04
Allow excluding certain source files from mapping

rodrigovilina:
**THIS IS STILL MAINLY A DRAFT**

A proposal on filtering source files as per suggested on the example in issue #118.

As I suggested there, I think it would be more granular and cleaner to implement if this exclusion was at each strategy level rather than at the `MapGenerator` level. Also each strategy could be dealt with one at a time, making review easier.

I'll leave comments on what could be simplified with this approach and what parts are ugly and will be refactored once we decide further course of action.

Since this is just a draft I haven't made any test already; I tested it locally (only on coverage strategy) by vendoring the gem and testing on some code that actually called also vendored `rspec` gem. Test will be my starting point for any further advancement.

I'll expect some type of response before continuing development.

### DIFF
--- a/lib/crystalball/map_generator.rb
+++ b/lib/crystalball/map_generator.rb
@@ -12,8 +12,8 @@ module Crystalball
       # Registers Crystalball handlers to generate execution map during specs execution
       #
       # @param [Proc] block to configure MapGenerator and Register strategies
-      def start!(&block)
-        generator = new(&block)
+      def start!(exclude_sources: [], &block)
+        generator = new(exclude_sources: exclude_sources, &block)
 
         ::RSpec.configure do |c|
           c.before(:suite) { generator.start! }
@@ -25,8 +25,8 @@ module Crystalball
       end
     end
 
-    def initialize
-      @configuration = Configuration.new
+    def initialize(exclude_sources: [])
+      @configuration = Configuration.new(exclude_sources: exclude_sources)
       @configuration.commit = repo.gcommit('HEAD') if repo
       yield @configuration if block_given?
     end
--- a/lib/crystalball/map_generator/configuration.rb
+++ b/lib/crystalball/map_generator/configuration.rb
@@ -12,8 +12,10 @@ module Crystalball
       attr_accessor :commit, :version, :compact_map
 
       attr_reader :strategies
+      attr_reader :exclude_sources
 
-      def initialize
+      def initialize(exclude_sources: [])
+        @exclude_sources = exclude_sources
         @strategies = StrategiesCollection.new
         @compact_map = true
       end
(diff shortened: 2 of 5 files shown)


lib/crystalball/map_generator.rb
```diff
@@ -12,8 +12,8 @@ class << self
       # Registers Crystalball handlers to generate execution map during specs execution
       #
       # @param [Proc] block to configure MapGenerator and Register strategies
-      def start!(&block)
-        generator = new(&block)
+      def start!(exclude_sources: [], &block)
```
rodrigovilina, line 15: this file would be unchanged if done at strategy level.

lib/crystalball/map_generator/configuration.rb
```diff
@@ -12,8 +12,10 @@ class Configuration
       attr_accessor :commit, :version, :compact_map
 
       attr_reader :strategies
+      attr_reader :exclude_sources
```
rodrigovilina, line 15: this file would also not be modified if implemented at strategy level.

lib/crystalball/map_generator/helpers/path_filter.rb
```diff
@@ -18,6 +23,8 @@ def filter(paths)
           paths
             .select { |file_name| file_name.start_with?(root_path) }
             .map { |file_name| file_name.sub("#{root_path}/", '') }
+            .reject { |file_name| exclude_sources.any? do |pattern| pattern.match?(file_name) end }
```
rodrigovilina, line 26: This is ugly right now, I'll refactor into a method and make it type safe for both strings and regexp instances and collections.

spec/map_generator/configuration_spec.rb
```diff
@@ -30,7 +30,7 @@
   end
 
   describe '#register' do
-    let(:strategy) { double(after_register: true) }
+    let(:strategy) { double(after_register: true, 'exclude_sources=' => []) }
```
rodrigovilina, line 33: Just so tests pass for the draft. I'll take a look into it once I continue.

lib/crystalball/map_generator/helpers/path_filter.rb
```diff
@@ -6,6 +6,11 @@ module Helpers
       # Helper module to filter file paths
       module PathFilter
         attr_reader :root_path
+        attr_writer :exclude_sources
+
+        def exclude_sources
```
rodrigovilina, line 11: Just being extra careful here, will probably remove once I have tests.

pirj:
Thanks for the PR!
I suggest summoning @jaimerson @EPecherkin and @pluff as reviewers.

rodrigovilina:
I'm picking this up again!

hebertdougl:
Hi @vaporyhumo, do we have any news here?

rodrigovilina:
Actually no, I switched jobs (I'm at toptal now :laughing:) and don't need this anymore. I can give it a try if it useful for anyone else tho. We can pair on it too if you want. :ear:

PR #134, opened 2022-01-09
Update to Psych 4.0.x

marksiemers:
Psych.safe_load changed it's method signature and requires
keyword args for 'permitted_classes'

Update the calls to Psych.safe_load and add it as an explicit
gem dependency with a pessimistic lock on the version.

### DIFF
--- a/crystalball.gemspec
+++ b/crystalball.gemspec
@@ -41,11 +41,12 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'parser'
   spec.add_development_dependency 'pry'
   spec.add_development_dependency 'pry-byebug'
+  spec.add_development_dependency 'psych', '~> 4.0.0'
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.0"
   spec.add_development_dependency 'rubocop', ">= 0.56"
   spec.add_development_dependency 'rubocop-rspec'
   spec.add_development_dependency 'simplecov'
-  spec.add_development_dependency 'sqlite3', "~> 1.3.13"
+  spec.add_development_dependency 'sqlite3', "~> 1.4.2"
   spec.add_development_dependency 'yard'
 end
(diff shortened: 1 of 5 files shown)


crystalball.gemspec
```diff
@@ -41,11 +41,12 @@ Gem::Specification.new do |spec|
...
   spec.add_development_dependency 'pry'
   spec.add_development_dependency 'pry-byebug'
+  spec.add_development_dependency 'psych', '~> 4.0.0'
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.0"
   spec.add_development_dependency 'rubocop', ">= 0.56"
   spec.add_development_dependency 'rubocop-rspec'
   spec.add_development_dependency 'simplecov'
-  spec.add_development_dependency 'sqlite3', "~> 1.3.13"
+  spec.add_development_dependency 'sqlite3', "~> 1.4.2"
```
marksiemers, line 50: An earlier version of sqlite was not building on my Mac, this upgrade was to unblock me from running the setup and the tests.

lib/crystalball/map_storage/yaml_storage.rb
```diff
@@ -33,7 +33,7 @@ def read_files(path)
 
           paths.map do |file|
             metadata, *example_groups = file.read.split("---\n").reject(&:empty?).map do |yaml|
-              YAML.safe_load(yaml, [Symbol])
+              YAML.safe_load(yaml, permitted_classes: [Symbol])
```
marksiemers, line 36: This version: `YAML.safe_load(yaml, [Symbol])` was deprecated by Psych, then removed in version 4.

See commit here: https://github.com/ruby/psych/commit/0767227051dbddf1f949eef512c174deabf22891#diff-659eac8589abc82c9a0ab3699e4e4be4774d9c09c6c9934af5d9dae0d264439cR322

marksiemers (commented):
Thanks for providing this gem. We're looking to incorporate it with our pre-push git hooks, but need it to work with the latest versions of Psych and Ruby.

marksiemers:
@pluff or @jaimerson - Any chance this could be reviewed?

mculp:
👀

